PR #5, opened 2012-01-23
fixed another problem with new status-field

francwalter:
fixed $statusIds and $statusNames and changed funktion in the db-layers (get_status > get_status_ids)


ServiusHack:
Merged with commit 3f53b50bf583314c7d4c6da2b0440d679c36e0b0.

PR #85, opened 2013-01-05
FIX: rounding is always done on bills, CHG: make use of formatted duration

simonschaufi:
This patch fixes a bug and adds variables to the rendering:
fduration: formated time used for each row
fttltime: formated total time (used in the aggregated footer)

### DIFF
--- a/core/extensions/ki_invoice/print.php
+++ b/core/extensions/ki_invoice/print.php
@@ -27,35 +27,39 @@ include_once('TinyButStrong/tinyDoc.class.php');
  * returns true if activity is in the arrays
  *
  * @param $arrays
- * @return true if $activity is in the array
+ * @param $activity
+ * @return bool true if $activity is in the array
  * @author AA
  */
 function array_activity_exists($arrays, $activity) {
-   $index = 0;
-   foreach ($arrays as $array) {
-      if ( in_array($activity,$array) ) {
-          return $index;
-      }
-      $index++;
-   }
-   return -1;
+	$index = 0;
+	foreach ($arrays as $array) {
+		if ( in_array($activity,$array) ) {
+			return $index;
+		}
+		$index++;
+	}
+	return -1;
 }
 
-function RoundValue( $value, $prec ) {
-   $precision = $prec;
+/**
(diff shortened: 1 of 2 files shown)


simonschaufi:
open https://github.com/kimai/kimai/pull/85/files?w=1 to see without whitspace changes. i changed the indent to tabs.

core/extensions/ki_invoice/print.php
```diff
@@ -92,81 +96,103 @@ function RoundValue( $value, $prec ) {
...
+			'username'=>$userName,
+			'useralias'=>$userAlias
+		);
+	}
+	$time_index++;
 }
 
 $round = 0;
 // do we have to round the time ?
 if (isset($_REQUEST['round'])) {
```
ServiusHack, line 166: This should test for $_REQUEST['roundValue'].

ServiusHack:
Your pull request cannot be automatically merged and there is one bug in core/extensions/ki_invoice/print.php at line 166. Please fix both, then I will merge it.

core/extensions/ki_invoice/print.php
simonschaufi, line 166: roundValue is always set since its a dropdown value whereas the checkbox needs to be checked. So only a combination of both would make sense. Am i right?

ServiusHack, line 166: Ah yes, you're right, missed that. Leave it as it is.

simonschaufi, line 166: ok, can you merge it then please?

ServiusHack, line 166: No I can not. As commented below it can not be automatically merged. I've taken a look at your changes and it seems they are mainly in conflict with 9ae750a5a422926ad593dd9ac55a35f6e97480a4. Please first solve those conflicts, then I will merge your code.

simonschaufi:
ok, i don't have much experiences with pull requests so far so i merged the whole stuff and created a new branch. this pull request can be deleted.

ServiusHack:
Updated merge request has been merged.

PR #153, opened 2013-04-05
Fixes login-issues via JSON-API

heiglandreas:
When using the JSON-API the translations-Object is not set in a global
scope but inside a function and is therefore not globally available.

This Pull-Request adds the global-keyword to the translations-variable to set it globally no matter where the file is included

### DIFF
--- a/core/includes/basics.php
+++ b/core/includes/basics.php
@@ -113,6 +113,7 @@ $database = new Kimai_Database_Mysql($kga);
 $database->connect($kga['server_hostname'],$kga['server_database'],$kga['server_username'],$kga['server_password'],$kga['utf8'],$kga['server_type'] );
 Kimai_Registry::setDatabase($database);
 
+global $translations;
 $translations = new Translations($kga);
 if ($kga['language'] != 'en')
   $translations->load($kga['language']);


kevinpapst:
Hi! I do understand the pull request, but could you please check if it is possible to add the
global $translations;
to the function instead of adding it to the basics include?

heiglandreas:
Hi.

It is already defined `global` in the function. But as the basic.php is included inside another function it has to be declared `global` in basic.php as well.

Otherwise it would require a larger rewrite. The problem is, that you would have to pass on the `$translation`-variable from  `Zend_Json_Server::handle()` through `Zend_Server_Abstract` and `Kimai_Remote_Api` to `Kimai_Database_Mysql` - Seems to me a rather large rewrite of partly library-code which should not be touched.

Alternative would be to either use `Zend_Registry` to store the `$translation`-Instance or create a `Translations::getInstance()`\- method to implement a getter to an already instanciated Translation-Object. That would then be some kind of Singleton-implementation.

Personally I would favour the Registry-approach as it allows easier replacement of the Translation-Object at runtime. 

That would require something like the following in `basic.php`:

``` php
$translations = new Translations($kga);
Zend_Registry::set('translation', $translation);
```

And then something like the following in `Kimai_Database_Mysql`:

``` php
$translations = Zend_Registry::get('translation');
```

I'm not sure where else that would then be necessary! I imagine a grep for 'global $translations' should suffice

And Yes, I know it's just another way of setting `$translations` globally ;-)

kevinpapst:
Even though I don't like like registering variables globalyl it is the quickest way to fix this problem.
Thanks for your Pull request!

PR #172, opened 2013-05-15
add quick note, default location, default tracking number

claudegel:
Hope this time it's Ok. I've added a quick note button in timesheet tab. added a timesheet tab in preferences to set default location, checkbox for auto tracking number and checkbox for on/off quick note button in timesheet tab


kevinpapst:
I have to frustrate you again, you made the second "beginners mistake". Be advised: That happens almost everyone, I had to redo pull requests because of that as well!

You commited whitespace changes, again very hard to review. I know I am gonna get kicked for that ;-) but could you do that one more time, but with different IDE settings? 
If you check out the differences in your Pull request (at https://github.com/claudegel/kimai/commit/a3db7fd6129cb34a764c62701dd7681089d99b1b) there are more than 60% changes which are not relevant and make it a pain to review. GitHub cannot detect what exactly was changed and displays so many insertions and deletions which in reality didn't happen on purpose...

simonschaufi:
Simply adding ?w=1 to the end of the URL will cause GitHub to ignore whitespace for all the diffs on the page. 

https://github.com/blog/967-github-secrets

Kevin, you are actually introducing all the whitespaces at the end of the line which should/could all be removed.

claudegel:
Great I can see the difference with and without the ?w=1. 
Do I need to erase the branch quick-note and push it again or with the ?w=1 it's ok for this time.

also what can I do in my IDE to avoid that problem in the futur, I'm using zend studio 10. Or can I just change my .git/config to have: url = https://github.com/claudegel/kimai.git?w=1

claudegel:
Sorry did I close/cancel the pull request by mistake

simonschaufi:
nop, the setting in the git config file wont change anything. you need to set that in the zend studio settings. i think the setting is called remove trailing whitespace or so. i used eclipse for a while and its almost the same.

kevinpapst:
Its an IDE setting, as simonschaufi wrote, search for "whitespace" in the ZS settings and disable it.

Adding the w=1 param helped a lot, thanks for that hint! 
But it still occurs to me that you mixed up several things: Please sent a Pull Request that really has one purpose. You incorporated some fixes (which is great btw) but it is disturbing when you try to understand the code changes. 

If you would use several commits that would allow us to cherry pick single files (which I personally don't like out of lazyness) but at least its better than mixing up several changes in on commit&pull request.

I need to repeat myself: Please don't get frustrated over this first review - the first ones are always hard ;-) We really appreciate your effort in improving Kimai!!! If you could try it a last time, that would be awesome, and use multiple branches, then we can merge the fixes directly and review the "big" change when there is more time.

claudegel:
Ok I was just going too fast. There is three changes so I will submit one for quick note (the big one), one for default location and one for default tracking number. I will also send one for the small fix and typo
Default location and tracking number are based on changes in the quick note as I've set a new tab in preference to have a checkbox to hide or show the quick note button on the timesheet tab. I'm using the same preference tab to add the other two changes as that way they are set as on/off feature and I will use that same preference tab in futur to add other  features in the timesheet tab.

in ZS I've remove the check for remove trailing white space. Thank you for your patience too

It took me 20 years to understand my wife so I think this time i will go faster ;-)

claudegel:
Le Wed, 15 May 2013 07:55:20 -0700,
Kevin Papst notifications@github.com a écrit :

> If you would use several commits that would allow us to cherry pick
> single files (which I personally don't like out of lazyness) but at
> least its better than mixing up several changes in on pull request.
> 
> I need to repeat myself: Please don't get frustrated over this first
> review - the first ones are always hard ;-) We really appreciate your
> effort in improving Kimai!!! If you could try it a last time, that
> would be awesome, and use multiple branches, then we can merge the
> fixes directly and review the "big" change when there is more time.

If the last pull request is ok I will do my changes for default
location and auto tracking number but for these two branch they are
based on changes made in the quick note branch. Can I create a new
branch below the quick note branch or wait for the approuval of the
quick note to proceed with those two addon.

## Thanks,

Claude Gélinas agr.
Phyto Ressources Inc

PR #174, opened 2013-05-15
Devel/quick note

claudegel:
Hope this time it's ok. I've updated only the files for quick note. If it is OK for you I think we should remove the comment and location in the extended tab of the edit timesheet floaters. This way I think that for security the note can be accessed by the user but in the edit timesheet floaters there are thing that should be accessed by the admin only.


kevinpapst:
We have to wait with the merge, as we currently have a feature freeze. We want to bring a new stable release and currently only do bugfixes or minor improvements (like your log message translation). 
Your pull request is therefor delayed until we released a new version.

PR #177, opened 2013-05-16
Delete log messsage configurable with language module

claudegel:
A small one to make the delete log message to be configurable in the language module.
The en and fr files already updated for that purpose (see end of files)

### DIFF
--- a/core/extensions/ext_debug/processor.php
+++ b/core/extensions/ext_debug/processor.php
@@ -73,8 +73,7 @@ switch ($axAction) {
             $logdatei=fopen(WEBROOT."temporary/logfile.txt","w");
             fwrite($logdatei,"");
             fclose($logdatei);
-            echo "Log file erased";
-            //echo $theData;
+            echo $kga['lang']['log_delete'];
         } else {
             die();
         }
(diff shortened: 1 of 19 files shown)


kevinpapst:
We need the translation in every language file. If you can't translate it, just add the key there.

And please: no more whitespace changes in the future. I do agree that re-formatting should be done ... but only if the team discussed first in which way - and then we should do it with one huge commit (or file-by-file) - but without logic changes in between.

ServiusHack:
Since quite some time it's not necessary to have the translation key in every translation file. First the English translation is loaded (in translations.class.php). Then this is merged with the language specified in autoconf.php (in basics.php). Finally this is merged with the language specified in the admin tab and the user preference (in Mysql.php). If one translation file doesn't contain a translation key then the English translation (or any other that overwrote it) is used.

kevinpapst:
True, but it makes it harder for people to translate everything to their language :)

claudegel:
Ok I'll add in each language. With google translation it's possible to get something basic. I will also update the fr.php as I found that many key were missing.

claudegel:
how can I delete a branch on github once it have been merged like my bugfix branch. I can delete it locally but can<t find out how to do it on github

kevinpapst:
The command is "git push origin --delete yourbranchname".

PR #180, opened 2013-05-18
Composer / phpUnit

kevinpapst:
This pull request adds a simple composer.json, which will install phpunit in the vendor directory.It shall be used for including external libraries and autoloading in the future.

Also I added a phpunit.xml and bootstrap.php to allow unit testing in the future.

The first unit test is actually not a real unit test, but searches for deprecated translations in all language files. 

The pull request includes several changes in the language files, which caused the unit test to fail - so I removed all keys, which are not required any longer. 

English translation contained the "extension" array twice, merged them as well.


kevinpapst:
Please don't merge, I just want some feedback here.

ServiusHack:
I don't really have a strong opinion on this. It's good to finally add some unit tests and composer seems like a nice solution. Time will show of how much use it is.

If you want we can merge this.

kevinpapst:
I will close this one, but we should still start using composer and unittests in the refactoring branch!

PR #183, opened 2013-05-20
do not activate demo extension by default

kevinpapst:
This pull requests contains a simple rename of the /extensions/ki_demoextension to /extensions/#ki_demoextension so it will not be loaded after a fresh installation or update


kevinpapst:
The other option would be to deactivate the permission even for admins. Either way, the demo extension should not be visible by default.

ServiusHack:
I'd much rather prefer using the permissions system. Enabling and disabling extensions should not require renaming folders on the server.

kevinpapst:
Definitely at your side with the permission system! 

But the demo extension is useless for 99% of all kimai users. I am not sure if we should even add it to the default installation. I'd rather delete it or keep it renamed. This way really interested developers can still use it and no one else should have ever have to wonder what it is and why its there.

ServiusHack:
But there would still be a permission for it. And enabling the permissions won't give access to the extensions. It must also be renamed. This seems more confusing to me.

Either we use the permissions system to disable the extension or there must be a general solution to installing and uninstalling extensions (copying files, adding the permissions).

kevinpapst:
The demo extension should not be part of the default kimai installation => it looks unprofessional!
The debug extension should be disabled by default => think for the user, not for us

Yes, we need to improve the extension system. And I think the permission for extensions should be stored in a own table "extension_permission" instead of being a column in globalRoles table with the columns:
name (string matching the directory name)
groupId
acces (boolean)

Otherwise we cannot easily add new extensions (I don't think adding a column is somehow an option).

EverPresent:
So this might be an ignorant question but what's wrong with adding an on/off switch for each extension into the configuration table?

It doesn't solve the problem of needing to add a column to the globalroles for the permissions info, but if you're trying to enable or disable an extension for the entire install then this seems the right place to do it.

ServiusHack:
The on/off switch would basically be what the rename is now. A global on/off is probably never what you want. At least for now, the demo extension should not be part of the release and the debug extension should be visible only to admins. The latter because it gives helpful information for bug reports (the log, actually never thought of that, I always told people to look at the file on the server) and maybe other things. I've created #203 to track this discussion.

As for this issue: I will merge the pull request so the demo extension won't show up by default. This solves the issue for now / the next release.

PR #188, opened 2013-05-24
fixed display error in tabular PDF-export

Sorato95:
Hey,

Here is the Fix for my issue: #167


ServiusHack:
Thank you for this fix.

PR #196, opened 2013-06-05
Sort invoice line items alphabetically by activity

### LINKED ISSUE #187: Order tasks alphabetically on invoice
There is a certain order in my project tasks, for instance:

01 Design
02 Build
03 Test
04 Deliver

I would like to have that exact order on my invoice (with sum hours), but at the moment the order is different. My request for change is to order the project tasks on the invoice alphabetically.
  
--
Hi vissersj,

I see that this can be of use. Please go to vissersj https://github.com/kimai/kimai/issues and file a feature request there.

Best Regards,
Severin

EverPresent:
I added a simple call to usort to fix #187

### DIFF
--- a/core/extensions/ki_invoice/print.php
+++ b/core/extensions/ki_invoice/print.php
@@ -156,6 +156,12 @@ while ($time_index < $timeArrayCounter) {
 	$time_index++;
 }
 
+// Sort alphabetically by activity.
+usort($invoiceArray, function($a, $b)
+{
+    return strcmp($a['desc'], $b['desc']);
+});
+
 $round = 0;
 // do we have to round the time ?
 if (isset($_REQUEST['round'])) {


kevinpapst:
I am not sure about this one here. How are they ordered currently?
We would change all invoices for all users with this commit.

Without reviewing the current code, I would say that this should be configurable through the UI for every invoice.

EverPresent:
I don’t believe there was a sort on this at all so I _think_ it was just displaying the list in whatever order the DB returned them.  But it’s been some time since I looked.

EverPresent:
Yeah.  Looking back I think I have to agree with you, Kevin.  The invoice print code has changed _considerably_ since this patch, it should also probably have a UI setting.  This should be re-coded.  I'll see if I have time in the next few weeks to update my installation and do it.

kevinpapst:
Ok, then I will close this pull request. Please send in a new one :)

PR #211, opened 2013-07-13
fixed a small syntax bug

vergissberlin:
Its just a small bug in line 46, but may be that does not work.
Thank you!


kevinpapst:
Thanks for finding and fixing this!

Next time it would be awesome not to push whitespace changes ;-)

PR #270, opened 2013-09-29
Redirect to installer when autoconf.php is missing. (closes #268)

ServiusHack:
Up until now an error message about the missing file was shown to the
user. In @4bc1c11 that file was removed from the repository thus always
giving an error to new users.

Instead of showing an error the user could immediately be redirected to
the installer. A missing autoconf.php now most likely means a plain new
installation.


kevinpapst:
We really need to include that and release a new beta.

PR #281, opened 2013-11-19
Increase width of nordic time-display because it's too small

hashworks:
For some dates the width set for the display in the skin nordic is too small:

screenshot_1

I'm using Chrome, didn't tested it in other browsers.

BTW: Is anyone using nordic anyway? I would include a skin like joyweb instead.


kevinpapst:
> I would include a skin like joyweb instead.

You would? Or you would like that someone does it? ;-)

PR #282, opened 2013-11-19
Fix php warnings on some systems

hashworks:
Commit a0ae103:
When starting the timer, sometimes an array is used in mysql_real_escape_string, 
which results in empty string and a warning:
"Warning: mysql_real_escape_string() expects parameter 1 to be string, array given in /var/www/zeiten_multiple/live/core/libraries/mysql.class.php on line 1579‏"
Nothing dangerous. Also I wasn't able to reproduce the warning, therefore I don't know why it gets an array there - so I applied a hotfix, resulting in the same result without the warning. Added a TODO comment.

Commit 2737f23:
This is slightly more important, since warnings in the ajax response kills the handling of them.
When using certain settings some values aren't used, therefore they generate a warning because the variable doesn't exist. I added various isset()-querys.

### DIFF
--- a/core/libraries/Kimai/Database/Mysql.php
+++ b/core/libraries/Kimai/Database/Mysql.php
@@ -1799,46 +1799,47 @@ class Kimai_Database_Mysql extends Kimai_Database_Abstract {
       return $this->conn->Query($query);
   }
 
- /**
-  * create time sheet entry
-  *
-  * @param integer $id    ID of record
-  * @param integer $data  array with record data
-  * @author th
-  */
-  public function timeEntry_create($data) {
-      $data = $this->clean_data($data);
+	/**
+	 * create time sheet entry
+	 *
+	 * @param integer $id    ID of record
+	 * @param integer $data  array with record data
+	 * @author th
+	 */
+	public function timeEntry_create($data) {
+		$data = $this->clean_data($data);
 
-      $values ['location']     =   MySQL::SQLValue( $data ['location'] );
-      $values ['comment']      =   MySQL::SQLValue( $data ['comment'] );
-      $values ['description']      =   MySQL::SQLValue( $data ['description'] );
-      if ($data ['trackingNumber'] == '')
-        $values ['trackingNumber'] = 'NULL';
-      else
-        $values ['trackingNumber'] =   MySQL::SQLValue( $data ['trackingNumber'] );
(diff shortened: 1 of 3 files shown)


kevinpapst:
Could you add one line with a log entry to your commit https://github.com/HashWorks/kimai/commit/a0ae10392bd79e8586c2bd56eb8f0d885a131ee5 and I am happy to merge it!

core/libraries/Kimai/Database/Mysql.php
```diff
@@ -1799,46 +1799,47 @@ public function timeEntry_delete($id) {
...
+	 *
+	 * @param integer $id    ID of record
+	 * @param integer $data  array with record data
+	 * @author th
+	 */
+	public function timeEntry_create($data) {
+		$data = $this->clean_data($data);
+
+		if (isset($data['comment'])) { $values ['comment']           = MySQL::SQLValue( $data ['comment'] ); }
+		if (isset($data['isDesignWork'])) { $values ['isDesignWork'] = MySQL::SQLValue( $data ['isDesignWork'] ); }
```
kevinpapst: What is "isDesignWork" and what happened to the "location" field?

hashworks: Ah, sorry - that's an internal variable, shouldn't be in here.
Seems like I forgot the location field.
9c48d7a

ServiusHack:
I'm sorry but we can't merge this pull request. There are several issues I see:
1. The main reason for this pull request seems to be that warnings lead to malformed AJAX responses. But this is only fixed for very specific cases. You probably want 0cf3ca658a2353ac86afc6eae94a7ab3bba87f71 to disable `display_errors` which is always a good thing in production. This absolutely and for all cases solves the issue of warnings corrupting the returned data.
2. You make every field of a timesheet entry optional. This won't work. A timesheet entry must have a userID, projectID and a lot more. Wrapping these in `if(isset(...` will hide bugs where this information is missing.
3. If under certain settings some fields are left out only these fields should be made optional.
4. The change in mysql.class.php suppresses a warning if something goes wrong but at least replaces it by a log entry. I just think the logging itself will not help much in debugging the issue if the value is not something you immediately recognize. Say it is `array()` then you won't have gained any insight. It should also log the backtrace so we know where the call came from. But should this even be part of a released version? Isn't it just of interest to developers until the bug is fixed?

simonschaufi:
closing with "not reproduceable" and no followup PR.

PR #297, opened 2014-01-25
added new description field to timesheet extension of the standard theme

mfriedr2:
hi, I pushed some  new features regarding the new description field and the standard theme. 
1. show the new description field as shortened column within timesheet, and toggle it by showtrackingnumber-flag (this is easier to use if there is only one checkbox for details in timesheetextension, at the moment i fitted languages en and de only)
2. show user alias if present instead of username only in timesheet extension - what else should be alias for?!
3. Resize Task Picker Status Area (upper right corner) to fit longer verbings in FF and IE. it is annoying if one can see only customer and project but not the task in the field, since german language headings need two lines of text. easier might it be, to resize the field in total. 
   TODO: the background image (g3_buzzer_display) should be resized, too, by 2px more in height and 20-30px more width, I don't have the psd ...

### DIFF
--- a/core/extensions/ki_timesheets/css/tables.css.php
+++ b/core/extensions/ki_timesheets/css/tables.css.php
@@ -115,6 +115,12 @@ div.ki_timesheet>div#timeSheet>div#timeSheetTable>table>tbody>tr>td.username {
     width:40px;
 }
 
+#timeSheet_head td.description,
+#timeSheet td.description
+{
+    width:450px;
+}
+
 #timeSheet_head td.from,
 #timeSheet_head td.to,
 #timeSheet td.from,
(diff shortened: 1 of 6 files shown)


kevinpapst:
Hi! Sorry, but you did a mistake in your pull request, that occurs very often when you use Git for the first times. You don't want to know how often it happened to me before I realized the problem ;-)

The problem is your commit https://github.com/mfriedr2/kimai/commit/8954d1da350124321ce22adf22c08018b87b8d21

> Merge remote branch 'upstream/master' into mfkimai

That completely rewrites our branch history, I can't accept that pull request therefor. Please ask Google about that problem. Just a very quick info on how to fix it:
- Fast forward your master to the latest version
- Create a branch
- Apply the three commits like you did here (very good that you did not change whitespaces!)
- Push that branch and send a new Pull request

Thanks for improving Kimai!

PR #300, opened 2014-02-01
New permission to hide users in same group.

ServiusHack:
This permission is very helpful for simple users who should not be
allowed to see any other users in their group.


jakubczaplicki:
Is there a workaround for this issue (for PHP 5.3) ?

kevinpapst:
I submitted a possible fix (cannot test it, as I don't have a PHP 5.3) some minutes ago. 
Please let me know, if that fixes your problem.

PR #301, opened 2014-02-01
Hide user list when unusable for filtering.

ServiusHack:
The user list is hidden when it is empty or contains just one user. In
both cases it is of no use to the currently logged in user.


azimjs:
The user list is empty even when i have other users with the same group ! help needed

PR #323, opened 2014-02-28
added new description field to timesheet extension of the standard theme

mfriedr2:
hi, I pushed some new features regarding the new description field and the standard theme.

 1   show the new description field as shortened column within timesheet, and toggle it by showtrackingnumber-flag (this is easier to use if there is only one checkbox for details in timesheetextension, at the moment i fitted languages en and de only)
 2  show user alias if present instead of username only in timesheet extension - what else should be alias for?!
 3   Resize Task Picker Status Area (upper right corner) to fit longer verbings in FF and IE. it is annoying if one can see only customer and project but not the task in the field, since german language headings need two lines of text. easier might it be, to resize the field in total. TODO: the background image (g3_buzzer_display) should be resized, too, by 2px more in height and 20-30px more width, I don't have the psd ...

this are the same changes as in pull request #297 that was closed but not merged because of a "Merge remote branch 'upstream/master' into mybranch", i.e. my false previous commit mfriedr2@8954d1d (8954d1da350124321ce22adf22c08018b87b8d21)

### DIFF
--- a/core/extensions/ki_timesheets/css/tables.css.php
+++ b/core/extensions/ki_timesheets/css/tables.css.php
@@ -115,6 +115,12 @@ div.ki_timesheet>div#timeSheet>div#timeSheetTable>table>tbody>tr>td.username {
     width:40px;
 }
 
+#timeSheet_head td.description,
+#timeSheet td.description
+{
+    width:450px;
+}
+
 #timeSheet_head td.from,
 #timeSheet_head td.to,
 #timeSheet td.from,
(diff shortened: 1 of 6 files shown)


kevinpapst:
I'm curious: who is "Root Account"?

mfriedr2:
very sorry, thats me myself with su root on my server.

Kevin Papst, 28.02.2014 16:47:

kevinpapst:
Okay ;-)

But now I can't merge the pull request automatically. Could you probably do the merging (as I don't know your code) and update this pull request?

PR #324, opened 2014-02-28
Rewrite of Kimai skins

kevinpapst:
This is a huge rewrite of the Kimai skin system.
Each skin now has complete control of every aspect of the rendered page, can (but most not) overwrite the HTML structure or only single viewhelper.

A new Skin "flat" (based on Twitter Bootstrap) was added, which is currently in beta state, as not all components were yet adapted (and it was only tested by myself during development).


kevinpapst:
Closing, as this was merged into the upcoming version 0.9.4

PR #338, opened 2014-03-02
Added rudimentary password reset functionality. (refs #81)

ServiusHack:


core/libraries/Kimai/Auth/Kimai.php
```diff
@@ -45,4 +45,53 @@ public function authenticate($username, $password, &$userId)
...
 
+    public function forgotPassword($name)
+    {
+      $kga      = $this->getKga();
+      $database = $this->getDatabase();
+
+      $id = $database->user_name2id($name);
+
+      $user = $database->user_get_data($id);
+      $passwordResetHash = str_shuffle(MD5(microtime()));
```
kevinpapst, line 56: I would use the current users password hash from the database, combined with microtime():
Something like:
`$passwordResetHash = str_shuffle(MD5($user['password'] . microtime()));`
The current implementation is predictable and therefor a security risk.

ServiusHack, line 56: `str_shuffle` introduces randomness and thus makes the result unpredictable. Assuming `str_shuffle` does not prevent predictability (and `MD5` and `microtime` are predictable, as they are) a brute force attack on the password reset hash can reveal the original password. Since the hash is available this can be done very rapidly. If you still see a security risk feel free to elaborate. You may do so via mail, if you want.

kevinpapst, line 56: The password hash itself (I assume $user['password'] is not a plain password but a salted hash) should be "secure" already. As long as no one knows the salt, brute forcing should already be quite a job with a salt length of 32(?) character. Adding up the extra layer of md5 + microtime and str_shuffle around that hash ... you really think that this could be brute forced "very rapidly"? The amount of permutations a hacker needs to try in that scenario sounds almost unlimited to me, but I have to admit that I am neither a security nor mathematic expert.

But at the end I might be wrong just from the point of view that there is no big security risk in a password reset function. Access to email account is required as well ... so I don't mind if you just ignore my comment ;-)

PR #368, opened 2014-04-30
Handle wrong database credentials ...

kevinpapst:
... or empty autoconf.php

Which resulted in hundreds of log entries and finally a white page. 
The underlying error was hard to spot (especially for non techies)

Happened for example with an empty autoconf.php which was existing in the beta version - see #349


kevinpapst:
@ServiusHack Do you see anything that could cause troubles? I tested and the installer redirect happens if autoconf.php does not exist. If it exists and the database connection is broken, the die() takes place.

ServiusHack:
Looks good. But we must also make sure not to include autoconf.php in a released version.

kevinpapst:
Thats why I deleted it from the repository and added it to .gitignore ;-)

PR #378, opened 2014-05-06
Implement new password hashing scheme.

### LINKED ISSUE #355: Better storage of passwords in DB - correct usage of salt/one-way-function
Currently _kimai_ stores passwords in the DB in the following way:
`data = md5($kga['password_salt'].$password.$kga['password_salt'])`

Since `$kga['password_salt']` is a site-static value, this is not a _salt_ in a cryptographic way, it's closest to the concept of _pepper_ instead.

While also a _pepper_ may have some security enhancing effects, it's not a suitable replacement for a _salt_. _Salts_ need to be assigned per user and unique in the "salt-space" of a site.

It is also bad practice to use hash-functions as _md5_ and the _SHA_-families for storing passwords. Those are designed to be fast, which ultimately aids an attacker, but isn't really needed when authenticating an user. It is better to use password encryption schemes which are designed for this special purpose.

The PHP FAQ contains a nice article[1] about this issue.
Furthermore, PHP 5.5 contains a high-level password handling API, addressing this issues.[2]

What I propose:
- Create 2 password handling functions for encoding and verifying passwords. This leads to a single point in the code where password handling is done. Currently there are 11 lines of code all over the place.
- Use a per user salt. The `password` field in the DB is already wide enough to hold all necessary data.
- Use a secure password storage scheme instead of `md5()`. This needs a bit of investigation of compatible PHP versions. There is a backport[3] of the new password API to PHP 5.3.7. Debian Squeeze for example just has 5.3.3, thus breaking compatibility.

Overall, this is not an immediate security threat. However, in my opinion such things should be done properly, especially when personal data of users are involved.

[1] http://php.net/manual/en/faq.passwords.php
[2] http://php.net/manual/en/ref.password.php
[3] https://github.com/ircmaxell/password_compat

scolytus:
This patch resolves #355.
# Features

The PHP built in `crypt()` function is used to encode the plaintext passwords. As the algorithm `$5$` was chosen because of compatibility reasons. This algorithm uses SHA-256 as a basis. The better `$2y$` (based on bcrypt) cannot be used since it is broken on PHP versions prior to 5.3.7 and would lead to users unable to login when the PHP version is updated.

The structure of the three new methods `encode_password()`, `verify_password()` and `password_needs_reencoding()` is similar to the new PHP Password API and the compatibility API. Once the PHP 5.3 compatibility requirement is abandoned it should be easy to switch to a newer standard API.
# TODOs / Talking points
- Currently the `crypt()` function uses 5000 rounds. That's the default and wasn't questioned.
- The `authenticate()` method is now implemented in `Abstract.php`. That's because of both `Http.php` and `Kimai.php` using it. (Btw, the`Http.php` code was/is a mess) Such methods should not exist twice. It could be placed in an intermediate class. I did not deem that necessary, though.
- The `encode_password()` function is also present in `func.php`. I'm not sure whether the Auth object is present everywhere the `encode_password()` function is needed.

### DIFF
--- a/core/core/processor.php
+++ b/core/core/processor.php
@@ -92,7 +92,7 @@ switch ($axAction) {
         
         // If the password field is empty don't overwrite the old password.
         if (trim($_REQUEST['password']) != "") {
-            $userData['password'] = md5($kga['password_salt'].$_REQUEST['password'].$kga['password_salt']);
+            $userData['password'] = encode_password($_REQUEST['password']);
             $database->user_edit($kga['user']['userID'], $userData);
         }
         
@@ -230,7 +230,7 @@ switch ($axAction) {
         
               // If password field is empty dont overwrite the password.
               if (isset($_REQUEST['password']) && $_REQUEST['password'] != "") {
-                $data['password'] = md5($kga['password_salt'].$_REQUEST['password'].$kga['password_salt']);
+                $data['password'] = encode_password($_REQUEST['password']);
               }
               if (isset($_REQUEST['no_password']) && $_REQUEST['no_password']) {
                 $data['password'] = '';
--- a/core/extensions/ki_adminpanel/processor.php
+++ b/core/extensions/ki_adminpanel/processor.php
@@ -384,7 +384,7 @@ switch ($axAction)
                 $userData['rate'] = str_replace($kga['conf']['decimalSeparator'],'.',$_REQUEST['rate']);
 		// if password field is empty => password unchanged (not overwritten with "")
 		if ($_REQUEST['password'] != "") {
-			$userData['password'] = md5($kga['password_salt'] . $_REQUEST['password'] . $kga['password_salt']);
+			$userData['password'] = encode_password($_REQUEST['password']);
 		}
 
                 $oldGroups = $database->getGroupMemberships($id);
--- a/core/includes/func.php
+++ b/core/includes/func.php
@@ -529,3 +529,17 @@ function coreObjectActionAllowed($objectTypeName, $action) {
 
   return false;
 }
+
+/**
+ * Encode a provided password as we need it to store in the DB.
+ *
+ * @param $password the password string to encode
+ * @return the encoded password string
+ */
+function encode_password($password) {
+  // use crypt with SHA-256 and 5000 rounds. As the actual salt we take the
+  // first 16 chars from a MD5 string.
+  $salt = '$5$rounds=5000$' . substr(md5(microtime()), 0, 16);
+  $pepper = $kga['password_salt'];
+  return crypt($pepper . $password, $salt);
+}
--- a/core/index.php
+++ b/core/index.php
@@ -147,7 +147,7 @@ switch($_REQUEST['a'])
 
         if ($is_customer) {
           // perform login of customer
-          $passCrypt = md5($kga['password_salt'].$password.$kga['password_salt']);
+          $passCrypt = encode_password($password);
           $id = $database->customer_nameToID($name);
           $data = $database->customer_get_data($id);
 
--- a/core/installer/install.php
+++ b/core/installer/install.php
@@ -312,7 +312,7 @@ exec_query($query);
 
 
 // ADMIN USER
-$adminPassword =  md5($kga['password_salt'].'changeme'.$kga['password_salt']);
+$adminPassword =  encode_password('changeme');
 $query="INSERT INTO `${p}users` (`userID`,`name`,`mail`,`password`, `globalRoleID` ) VALUES ('$randomAdminID','admin','admin@yourwebspace.de','$adminPassword',1);";
 exec_query($query);
 
--- a/core/libraries/Kimai/Auth/Abstract.php
+++ b/core/libraries/Kimai/Auth/Abstract.php
@@ -122,7 +122,47 @@ abstract class Kimai_Auth_Abstract
      * @param int $userId is set to the id of the user in Kimai. If none exists it will be <code>false</code>
      * @return boolean either <code>true</code> if the credentials were correct, or <code>false</code> otherwise
      **/
-    abstract public function authenticate($username, $plainPassword, &$userId);
+    public function authenticate($username, $plainPassword, &$userId)
+    {
+        $database = $this->getDatabase();
+
+        $id = $database->user_name2id($username);
+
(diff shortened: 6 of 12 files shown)


kevinpapst:
We need to stick to PHP 5.3 at least for Kimai 0.9.3. So we should think about merging that request with the best working solution for 0.9.4 where we can switch to PHP 5.4 (PHP 5.5 is definitely out of scope, due to the lack of upgraded PHP version by most of the hosters out there). 

Yes, there is still a lot of messy code in Kimai (we took over the project some while ago and speaking for myself: I only updated the code that I touched). You don't wanna know how it looked when I saw Kimai the first time ;-) I am so glad to get some qualified feedback and pull requests!

Like I said in the other pull request: I need some days for evaluation, as some other projects are more urgent right now...

scolytus:
No need to hurry, I just had a few minutes... And I'd suggest to prioritize **this** request over the other since it is an important topic IMHO. The other works for me as is and I just wanted to share it.

Regarding Versions: It's just bad. For me it is important to be compatible with Debian Squeeze, meaning PHP 5.3.3 with some backports. What really sucks is that `crypt()` behaves different starting from `5.3.7`, so a poin-point-release breaks compatibility. :-1:  So I personally think that being usable on Squeeze while it's maintained is a good thing, also for 0.9.4.

Maintaining compatibility was my major concern when writing the code, I don't like to see users locked out because of a PHP update.

martinvahi:
Well, I hate to rant and brag, or may be I love it, depends on the point of view :-> but it is a fact that hash functions and passwords is quite a [tedious topic](http://longterm.softf1.com/specifications/hash_functions/plaice_hash_function/).

I'm not a security expert, but a preliminary, dirty, workaround to many problems might be to explicitly SPECIFY CONSTRAINTS that Kimai (or whatever other Your loved web application) is supported ONLY on x86 and AMD64, e.g. on hardware, where both, [byte endianness](https://en.wikipedia.org/wiki/Endianness) and [bit endianness](http://en.wikipedia.org/wiki/Bit_numbering) is "equal to AMD64", e.g. little endian. That should support all AMD64 CPU-s, regardless of manufacturer, all x86 CPU-s, regardless of manufacturer, and it also gives the ARM also a chance, because the ARM is said to be bi-endian and rumored to be "in little endian mode on Android". High-end Intel CPU-s (read: ridiculously expensive Enterprise hardware), IBM custom machines and European Space Agency financed, open source,  [LEON](https://en.wikipedia.org/wiki/LEON) will, unfortunately, not be an option, because they are big-endian.

I'm not sure, if it helps, but as little as I currently understand the [length extension attack](https://en.wikipedia.org/wiki/Length_extension_attack), a workaround might be

```
hash(secret_1+the_text_that_might_be_publickly_known_to_attackers+secret_2)
```

where secret_1 and secret_2 might be pieces of the password or secret_1 and secret_2 might be even equal. To counteract the [rainbow table attack](https://en.wikipedia.org/wiki/Rainbow_tables), a proper solution might be:

```
patched_hash_type_1(secret,something_public_and_variable_and_very_long)=hash(
secret+something_public_and_variable_and_very_long+secret)
```

where the

```
something_public_and_variable_and_very_long=public_session_ID_that_has_at_least_40_characters
```

A third countermeasure is to reduce the probability of collisions. A collision is a situation, where

```
hash(text_1)==hash(text_2)
```

The countermeasure:

```
patched_hash_type_2(secret,long_and_public_1, long_and_public_2)=
=patched_hash_type_1(secret,long_and_public_1)+patched_hash_type_1(secret,long_and_public_2)
```

As I'm not a security expert and I probably miss most of the issues that are related to hash functions, but it seems to me that some hash function design deficiencies can be compensated by using hash functions wisely. 2 links: [BLAKE2](https://blake2.net/) and [something plain fun](https://www.youtube.com/watch?v=qKk8MHFLNNE#t=70).

Thank You for reading my comment. :-)

scolytus:
@martinvahi I'm not quite sure what you want to tell us by your comment. Have you read the patch? Can you elaborate a bit on which parts of the patch you refer with your comment?

martinvahi:
@scolytus An explanation by analogy is that in stead of telling someone, "turn the car right over here", it is more informing to tell, "if You drive straight forward, You might hit X, because ...,  If You turn left, You will likely loose tire Y, because ...,  If You turn right from the first road ahead, You will likely fall off the cliff, because ....,  but if You turn right at the second road ahead, You might just make it, even, if it's not sure that there are no surprises waiting there."

My comment was not about a specific patch. I do not remember, how or why I found this thread, but I do like the idea behind the Kimai project and since I have been working on hash function related code recently, I thought that I'll save You (and myself) a bit of trouble by pointing out a few things. Otherwise I might have to open a new feature request or bug report later, after You have applied your patches.

If You do not find my previous comment relevant, then I'm sorry for bringing noise to this discussion and it's OK to ignore my comment. Sorry. I was just trying to help.

kevinpapst:
We always appreciate code and input from the community! But I was also wondering what you wanted to say ;-) Yes, security is a very complicated field and we have to improve here a lot. I guess some patches would be better, to show us your idea in practice and not only in theory. 
I am a web developer and just have a basic understanding of security and would love to see your ideas the same as I like @scolytus input. IMO thats the real spirit of Open Source.

martinvahi:
@kevinpapst   :-)

scolytus:
Since there is some activity here again, @kevinpapst what do you need to take this further? What can I do to support you?

kevinpapst:
Thanks for questioning, its just missing time that stands in our way. We need to fix some bugs and then publish the current release candidate. As soon as that happens, I will have a look and either comment or merge your code directly!

scolytus:
Any news?

kevinpapst:
I am so sorry for not responding earlier! Still planning the release ... but time is such a fluent element^^

cljk:
I really like the idea about switching the hashing algorithm and abstracting the authentication methods - MD5 is dead.

If you want to use bcrypt and be compatible with older PHP - why not use a small lib?
like... 
https://gist.github.com/marcoarment/1053158

For me I feel safer using a hashing alg wich is designed for using with passwords (= slow) and not using a hash which is designed to be fast (SHA) and making it slow by using more iterations.

simonschaufi:
closing this as we won't do huge refactorings any more.

PR #382, opened 2014-05-07
Implement a new authentication method based on SSL client certificates.

scolytus:
I propose an additional authentication method based on SSL client certificates.

This patch shall serve as a quick and dirty proof of concept. Further discussion very welcome.

The basis for this implementation is my personnel setup where Apache denies the SSL handshake at all if the user does not provide a valid certificate.

### DIFF
--- a/core/libraries/Kimai/Auth/SSLCert.php
+++ b/core/libraries/Kimai/Auth/SSLCert.php
@@ -1,0 +1,161 @@
+<?php
+/**
+ * This file is part of
+ * Kimai - Open Source Time Tracking // http://www.kimai.org
+ * Copyright (C) 2014 Michael Gissing
+ *
+ * Kimai is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; Version 3, 29 June 2007
+ *
+ * Kimai is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Kimai; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * This class authenticates a Kimai user based on an SSL client certificate.
+ *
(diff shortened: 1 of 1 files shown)


kevinpapst:
Very interesting! And thanks for your qualified input and improvements, they are very welcome :+1: 

I need to discuss whether we merge all of your pull requests before the release or directly afterwards. Give me some days for evaluation and testing. 

Do you know if that code works on nginx the same way (just curious, because you are talking about Apache only)?

scolytus:
I'm talking Apache only because it's the only environment I currently use.

Basically the concept must be the same for nginx. A brief Google search also showed that people already did it. However, I can't guarantee that the code works as is. And I don't know how to configure nginx to provide the values in `$_SERVER`.

If there is an easy way for me to get hands on an already running nginx I could take a look.

scolytus:
I've looked into the nginx thing a bit and I think it won't work well.

The problem is that nginx does not provide the different fields of the certificate as `$_SEVER` values. It just provides the `DN` string which is not standard conform and hard/impossible to parse. (The OpenSSL guys strongly recommend not to use the function that nginx uses...) So I won't implement this in PHP.

As an alternative one could configure nginx to handover the raw certificate to PHP. This may have some performance implications since the data has to be copied to the environment on every call. I haven't yet looked into `openssl_x509_parse()` but it looks doable. So maybe if I have another spare hour and implement this as well - again - as a proof of concept.

kevinpapst:
Thanks for looking into it! I love and use nginx, but I think the vast amount of all installations still use Apache. I will not say no, but probably concentrating on the current request is enough. 
You said its a proof of concept, so if you don't see it as code that is production ready, probably you want to improve that instead?!? Just thinking out loud ;-)

cljk:
Why not integrate if it works on most (~80%) sites - on nginx it will just not login.

Client certificate check itself is done in the webserver - before PHP comes in... so no security concern here for me. If the certificate is not valid a client will not be able to use the application (if the server is configured correctly).

simonschaufi:
Hi @scolytus Could you write some documentation for it how to use it? Then I will be glad to merge it!
Please add it in the documentation repo!

scolytus:
@simonschaufi I wrote this more than 3 years ago, I don't use kimai any more and I won't contribute to it either.

simonschaufi:
closing this as we won't do huge refactorings any more. please use kimai 2

PR #384, opened 2014-05-14
=closes #383

paulsteffen:
- permission check in function "timesheetAccessAllowed" fixed to work with editLimit setting
- fixed one typo in german translation

### DIFF
--- a/core/extensions/ki_timesheets/processor.php
+++ b/core/extensions/ki_timesheets/processor.php
@@ -35,7 +35,7 @@ function timesheetAccessAllowed($entry, $action, &$errors) {
   }
 
 
-  if ($kga['conf']['editLimit'] != "-" && time()-$entry['end'] > $kga['conf']['editLimit']) {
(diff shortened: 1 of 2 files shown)


kevinpapst:
Could you please describe the problem, so I can reproduce it?

kevinpapst:
Ah, okay, I found the problem. Let me check your fix and then I will merge it!

paulsteffen:
If editLimit is set in the configuration it is impossible to start a new recording based on another one.
(Admin Panel -> Advanced -> check "No entries can be added or edited with en more than" and insert any value)
If you click on the "Record this recording again" button in front of any timesheet entry the recording does not start 

/extensions/ki_timesheets/processor.php line 96 ->  $timeSheetEntry['end'] = 0;
so the function "timesheetAccessAllowed" will always return false if  $kga['conf']['editLimit'] != '-'

kevinpapst:
:) Very good, thanks for finding and fixing!

PR #387, opened 2014-05-15
FEATURE: Add visible to getCustomers and add tasks to getProjects

waldi:
did i make it right?

### DIFF
--- a/core/includes/classes/remote.class.php
+++ b/core/includes/classes/remote.class.php
@@ -377,7 +377,7 @@ class Kimai_Remote_Api
      * @see 'reload_projects'
      * @return array|boolean
      */
-	public function getProjects($apiKey)
+	public function getProjects($apiKey, $includeTasks = false)
 	{
         if (!$this->init($apiKey, 'getProjects', true)) {
 			return $this->getAuthErrorResult();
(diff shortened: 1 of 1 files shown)


core/includes/classes/remote.class.php
```diff
@@ -394,6 +394,38 @@ public function getProjects($apiKey)
 		}
 
         if (count($projects) > 0) {
+
+            $tempProjects = array();
+            foreach($projects as $project) {
+                $tasks = array();
+
+                // @FIXME
```
kevinpapst: Why is there a FIXME?

kevinpapst:
Can you 1) explain the FIXME 2) fix it or 3) remove it?

waldi:
The FIXME is from the "getTasks"-Code. I didn't wrote it, neither don't i know why it's there.. should i still remove it?

kevinpapst:
Okay, I checked the current remote code and do see where it comes from...
Actually, I really don't like copy & pasting (extract method would be the correct thing here). 

What I would like to see here is:
An extracted method to fetch the tasks by projectID, which then is used in both the getTasks() and getProjects() methods. 

Also, the getProjects($apiKey) signature should be changed to include a second and optional parameter $includeTasks =>  getProjects($apiKey, $includeTasks = false)

There are some programs out there using the remote API. They might break from that change, as it is not backward compatible. Do you want to include these changes?

waldi:
ok. yeah i also thought about adding a extracted method. i'll do that.

yeah i'll do that change.

kevinpapst:
:+1: thanks

waldi:
ok i fixed it

Issue #284, opened 2013-11-21
In released beta and latest SVN build, invoice shows UNIX style dates...


alauppe:
...instead of strftim() formatted dates

I have the date format set to %m-%d everywhere. Before my upgrade (from 0.8 I believe) invoices had the proper date, but now, today for example, looks like 1385036389.

Thanks!

ServiusHack:
You were both right, the issue is still not fixed. I actually broke it again in f5b4790f9fe16d159adc6cc2a104cc1d5543a241 which was the only change between my original fix and now.

So with #458 I've performed the fix again. First, it must be "frm" and not "fmt". Second, there must be no formatting instructions for everything from the opening bracket to the closing one.

You can check for the formatting instructions by opening the odt file using some ZIP unpacker and then opening the contained content.xml file. Searching for "today" should show the whole formatting statement without any XML-Elements in between.

PR #396, opened 2014-06-03
Update print.php

jimmymarbles:
This should correct "today's date" for the custom invoices.  Issue #284


kevinpapst:
Okay, I'll have a look as soon as I can. Thanks for the patch!

ServiusHack:
This won't actually fix #284. It works for you because you format the time in PHP using the `date` function and giving it a format string. The result is some text which is not altered because the formatting through the [var.today...] syntax is not working.

We won't merge this pull request as it's not solving the issue at its root cause. I've created a new pull request to solve the issue: #458

PR #400, opened 2014-06-28
Cronjob to send customer a daily report via email

simonschaufi:
Timesheet items from the _previous day_ will be used.
If time spans reach the next day, the last entry will be cut off at midnight and the rest will be included in the following day.
Setup the cron like this:
5 0 \* \* \* /path/to/php /absolute/path/to/kimai/shellscripts/cron.php
this will run 0:05 every day.
For the cronjob to work, it must be activated for the customer and an email address must be set. Also the system administrator email address should be set. This will be the sender address.

TODO:
- [ ] make email subject configureable

IDEAS:
- cronjob extension
- weekly reports
- monthly reports
- Pivot (#314)

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -107,6 +107,9 @@ switch ($axAction) {
             }
         } else {
             $view->assign('timezone', $kga['timezone']);
+            $view->assign('customer', [
+                'cronJobActive' => false
+            ]);
         }
 
         $view->assign('timezones', timezoneList());
--- a/core/processor.php
+++ b/core/processor.php
@@ -253,6 +253,7 @@ switch ($axAction) {
                 $data['homepage'] = $_REQUEST['homepage'];
                 $data['visible'] = getRequestBool('visible');
                 $data['filter'] = $_REQUEST['customerFilter'];
+                $data['cron_job_active'] = getRequestBool('cronJobActive');
 
                 // If password field is empty don't overwrite the password.
                 if (isset($_REQUEST['password']) && $_REQUEST['password'] != "") {
--- a/extensions/ki_adminpanel/templates/scripts/customers.php
+++ b/extensions/ki_adminpanel/templates/scripts/customers.php
@@ -1,4 +1,4 @@
-<a href="#" onclick="floaterShow('floaters.php','add_edit_customer',0,0,450); $(this).blur(); return false;"><img src="<?php echo $this->skin('grfx/add.png'); ?>" width="22" height="16" alt="<?php echo $this->translate('new_customer')?>"></a> <?php echo $this->translate('new_customer')?>
+<a href="#" onclick="floaterShow('floaters.php','add_edit_customer',0,0,800); $(this).blur(); return false;"><img src="<?php echo $this->skin('grfx/add.png'); ?>" width="22" height="16" alt="<?php echo $this->translate('new_customer')?>"></a> <?php echo $this->translate('new_customer')?>
 <br/><br/>
 <table>
 	<thead>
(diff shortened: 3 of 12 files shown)


core/shellscripts/cron.php
```diff
@@ -0,0 +1,125 @@
...
+			$att->encoding = Zend_Mime::ENCODING_BASE64;
+			$att->filename = $fileName;
+			$att->id = md5($att->filename);
+
+			// Embed Images
+			$view->embed = array('logo' => 'cid:'.$att->id);
+
+			$output = $view->render('shellscripts/timeSheet.php');
+
+			#$mail->setBodyText(''); //TODO: no plain Text available yet
```
ServiusHack: Do you plan on implementing plain text for this mail? Maybe just add a second template file which generates plain text instead of HTML.

ServiusHack:
In `core/core/floaters.php`, `core/libraries/Zend/View/Helper/FormCheckbox.php` (why?) and `core/templates/scripts/floaters/add_edit_customer.php` I see major changes from whitespaces to tabs. Please revert those reformattings.

I really like this new feature. It would be great if I can merge it in after those few minor adjustments.

core/shellscripts/cron.php
simonschaufi: not for now. will come later if i have a good idea how the "layout" could look like.

ServiusHack:
We don't have any guidelines other than "keep consistent with what already exists". I agree that we should use some styleguide, but for now the situation is as it is. I don't like to merge in reformatting changes as people's opinion differ on what is the "right" way. Thus the situation does not really improve and only clutters up the history, diffs, blame, etc.

In any case, you should not change a library file without a really good reason. There is none for the `FormCheckbox.php` file. And when you're already reverting the formatting please do the same for the other files. In return I promise to put deciding on a styleguide as a top priority.

kevinpapst:
I double Severins opinion. There is no styleguide, but we can talk about creating one. 

But please no white space changes in a commit ... we get in big troubles when we have to merge them within several branches. We are already working on the next release and there are two big feature branches, which contain changes already.

simonschaufi:
Ok, but I still have one problem: when i use my IDE, how should i allign my code if everything looks like a big clutter? By the way, you can view a commit without white space changes by adding ?w=1 to the diff url to see the "real" changes but who cares about blame?
I will modify my patch even if it means a lot of work. I didn't want to touch any library, i just saw that the docheader was wrong and changed it but didn't realize it was in the Zend Framework ;)
What do you think about one big commit after the next release to align the code properly? If you do a git rebase, git should do its best to do its job.

kevinpapst:
To my experience, no one but everyone cares about blame ;-)
I know the w=1 switch, still doesn't help a lot with merging. We can do a big cleanup merge once we are at a stable and "no working in progress" point, sure. 

We live with that clutter since a long time, so literally its possible, even if it is annonying from time to time. Try to set your tab size to 4, that should help with the biggest pain.

kevinpapst:
Looks good, could you write a documentation chapter for the cronjob and its setup? Its this repo: https://github.com/kimai/documentation

Oh, and we cannot merge it before we released 0.9.3. Otherwise you could sent the pull request against https://github.com/kimai/kimai/tree/0.9.4 which would be easier (I believe there is some merging to do). We could merge that one directly.

kevinpapst:
Hey @simonschaufi - what about that PR?
Wanna change it and prepare it for develop?

simonschaufi:
there needs to be more done than just updating it to make it mergeable. right now the mail configurations are hard coded which should be set via admin panel or config file (to send via smtp)

kevinpapst:
So you want to keep it here in place as a reminder or do you want to close it and sent it in again, once your have rewritten it?

simonschaufi:
let's leave it for now.

PR #405, opened 2014-07-16
Pgsupport

albertlast:
Now with PostgreSQL support _dance_

Many things are done...
- first create a db wrapper mysqlw.class.php class name ist MySQL to get backwards compatiblity
- change the class name of mysql.class.php to MySQLNew (not relly new ...) but now with default sql_mode = ansi _yeah_ it's needed for comone sql language special for the install (alternative we can run mysql only in the moment of installing in ansi mode)
- vars.php edit that it now read the server_type from autoconfig
- install.php transform the create sql statement to a comone one, which pg and mysql unstand (the moste time) for this I change all tinyint to smallint -> 1byte to 2 bytes , remove the size information in "INT(x)", remove back ticks, added some function to handling the last differents
- \- one thing i can't transform the column comments, every one can do it, but got some differents way
  so I got to choose two install scripts or one without comments...
- the hole install process give you now the selection which db you want, when the extension for mysql and or pg are loaded
- http.php looks like a api this wars realy hardcoded to mysql_\* I change this, but don't know if this relly works... need phpunit test or something like this to test
- update.php I didn't touch there are so many things... for pg is update.php not needed because first release and mysql yeah big changes in the datatype ....

give me some input which modifaction is needed
this request need to be testet in a mysql environment!


albertlast:
I will try to get a cleaner commit with rebase

kevinpapst:
Thanks, very good idea ;-)

PR #407, opened 2014-08-15
Adds a highly configurable LDAP-Adapter

heiglandreas:
This PR replaces the currently available LDAP-Adapter with a highly configurable one. 

The new LDAP-Adapter can be used as authentication adapter for any LDAP-Server. Usage and Configuration are described in the acompanying LDAP_README.md file

Applying this LDAP-Adapter is a clear BC-Break as current configurations will not work any more. On the other hand, after an update the administrator has to reconfigure the LDAP-Adapter in any case as it has been replaced during the upgrade. Therefore that should not be an issue!

PR #413, opened 2014-09-15
Adds a highly configurable LDAP-Authenticator

heiglandreas:
Then documentation for this Authenticator can be found in the
Authenticator documentation.

This PR replaces #407

### DIFF
--- a/core/libraries/Kimai/Auth/Ldapadvanced.php
+++ b/core/libraries/Kimai/Auth/Ldapadvanced.php
@@ -1,0 +1,420 @@
+<?php
+
+/**
+ * Copyright (C) 2011 by Skaldrom Y. Sarg of oncode.info
+ * Copyright (C) 2014 by Andreas Heigl<andreas@heigl.org>
+ *
+ * This is free software. Use it however you want
+ *
+ * To activate this Authentication Adapter, add the following line to the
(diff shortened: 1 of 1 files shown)


kevinpapst:
Any chance the LDAP_OPT_PROTOCOL_VERSION needs to be configurable?

kevinpapst:
Oh any I would appreciate a new pull request with a version of this adapter, that will load a config file. 
From my perspective it is not very convenient, that a user has to create a new class and overwrite the base class properties in order to use your Adapter. 
Better create a config file with empty or some nice-default values that be used in combination with $authenticator = 'Ldapadvanced'; directly.

heiglandreas:
Is there any special way that config files can be loaded in Kimai? I didn't find one therefore I used the same way the LDAP-Authenticator has been configured. A config file would be much more convenient!

I can include the protocol-version-option in there as well.

kevinpapst:
Good question. I am not sure if we have a "default way" right now. Guess not ...

Maybe put it in https://github.com/kimai/kimai/tree/master/core/config as ldapadvanced.ini and use then use parse_ini() ?

kevinpapst:
Must be merged aftewards: https://github.com/kimai/documentation/pull/1

PR #414, opened 2014-09-16
Adds a testing framework

heiglandreas:
This PR adds some phpunit-files to make testing easier as well as adding some files to enable testing via TravisCI.

Tests can be placed inside the `test`-folder and a simple `phpunit` will then execute them. You might have to execute `composer dump-autoload` to get the autoloading right.

The Repository can also be added to travis-CI.org and on every commit the tests are executed on all PHP-Versions including hhvm. The Code-Coverage only includes the coverage of the tested files. All not tested files are not included!

### DIFF
--- a/.travis.install.sh
+++ b/.travis.install.sh
@@ -1,0 +1,11 @@
+set -x
+if [ "$TRAVIS_PHP_VERSION" = 'hhvm' ] || [ "$TRAVIS_PHP_VERSION" = 'hhvm-nightly' ] ; then
+    curl -sS https://getcomposer.org/installer > composer-installer.php
+    hhvm composer-installer.php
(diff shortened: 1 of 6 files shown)


kevinpapst:
I did something like this a while ago and then dumped the pull request ... just because I don't see a real chance that we can ever test the current  Kimai codebase properly. At least not the relevant parts with the business logic. The legacy code is just too weird. 

I can nevertheless merge the pull request, probably Severins refactoring will allow us to improve that part. But could you please change from test/ to tests/ and do a delete commit for the current test/ directory?!

heiglandreas:
My Idea was to just provide the framework so that everyone can add tests for new stuff if they want to. Otherwise tests will never happen to pop up ;-) 

But perhaps Severin should have a look at it too. 

Of course I will "move" test to tests.

kevinpapst:
What I totally appreciate :) Thanks for your support!!!!

PR #423, opened 2014-10-31
it.php

fbedussi:
Updated Italian translation


ServiusHack:
The file is in the root of the repository instead of in the core/langauge directory and it seems to be based on an old translation file. It has keys like knd and def_evtID. Please use the latest translation file core/language/it.php as a basis for your modifications and create a new pull request.

PR #428, opened 2014-11-26
Pgsupport clean

albertlast:
PostgreSQL Support for kimai

Use Mysql Class as Wrapper for both rdbms
use strict sql language which are run on both rdbms
extend the install wizard about pg support


cljk:
I like the idea of using lower case table names. Just had a problem migrating Kimai database from one system to another because od uppercase table names....

If the patch works, why not merge it?!

albertlast:
Biggest problem is that the patch is realy old (no auto merge), huge and
force strict sql (ansi mode) in mysql.

simonschaufi:
closing this as we won't do huge refactorings any more. please use kimai 2

PR #433, opened 2014-12-12
Auto link timesheet tracking number to timesheetentryID with on/off in p...

claudegel:
This is a small change to automatically link the new timesheetID number to the timesheet tracking number. I've put an on/off checkbox switch into the preferences

### DIFF
--- a/core/libraries/Kimai/Database/Mysql.php
+++ b/core/libraries/Kimai/Database/Mysql.php
@@ -1177,7 +1177,7 @@ class Kimai_Database_Mysql extends Kimai_Database_Abstract {
           $this->remove_rate($data['userID'], NULL, NULL);
         }
       }
-    
+
       return $data['userID'];
   }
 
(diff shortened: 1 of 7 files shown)


core/libraries/Kimai/Database/Mysql.php
```diff
@@ -4436,17 +4443,29 @@ public function membership_roles() {
...
+   * @return integer next auto_increment number
+   */
+   public function getNextAutoNumber($table){
+      if (! $this->conn->NextAutoNumber($table)){
+         $this->logLastError('Cannot get the auto number');
+         return false;
+      }else{
+         return $this->conn->NextAutoNumber($table);
+      }
+   }
```
simonschaufi: Personally I don't like patching external libraries. Extending this class and then putting this method in would be nicer.

claudegel:
we can create a file extend.php and add that function in.

class Kimai_Database_Extend extends Kimai_Database_Mysql {
  /**
 * get the next auto number
 * @param string $table
 * @return integer next auto_increment number
 */
 public function getNextAutoNumber($table){
      if (! $this->conn->NextAutoNumber($table)){
           $this->logLastError('Cannot get the auto number');
           return false;
      } 
     else
     {
           return $this->conn->NextAutoNumber($table);
      }
 }
}

simonschaufi:
oh sorry, i just see that this is Kimai's own library, not external ;) So this code can be included without probs.

simonschaufi:
Can you still explain shortly what this feature is all about? I don't really understand it from reading the code.

claudegel:
This code is to be able to get automatically a new tracking number when we start a new timesheet entry. It allow to use the timesheetID number as the tracking number. The tracking number field is filled automatically with the timesheetID number. That behavior is controled by a checkbox in the preference to allow the user to have the choice to use it or not.

simonschaufi:
I still don't understand why you need a tracking number. What's the difference between the timesheet id and the tracking number?

claudegel:
Hi, when you start a new timesheet entry you must fill (edit) the timesheet data in the floater. In the advanced tab you need to fill the location, the tracking number etc. For location I've submitted another patch to auto fill it with a default location because many time I always put the same location. For the tracking number you can put a PO number from your customer or just add a sequential number for which I use the timesheetID. This way I have two field automatically filled when starting a new timesheet. I suppose that the tracking number it to have a unique number to identify that specific timesheet.

claudegel:
So what next. Don't you think it is usefull to have an autoincrement flexibility for the tracking number ?

kevinpapst:

PR #435, opened 2014-12-12
Export to PDF in standard time

claudegel:
I've added the possibility to hide the decimal_time and total amount by clicking the title of each column in the export extention and to select only standard time instead of decimal time to export into PDF file

### DIFF
--- a/core/extensions/ki_export/base_export_pdf.php
+++ b/core/extensions/ki_export/base_export_pdf.php
@@ -1,7 +1,7 @@
 <?php
 
 
-class BasePDF extends TCPDF { 
+class BasePDF extends TCPDF {
 
   public $print_time;
   public $date_format;
@@ -14,7 +14,7 @@ class BasePDF extends TCPDF {
    */
   public function date($number) {
       return strftime($this->date_format,$number);
-  } 
+  }
 
   /**
    * Format a unix timestamp as a time string.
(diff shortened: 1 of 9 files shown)


simonschaufi:
@claudegel this patch removes a lot of whitespace which should not be done in a PR.
second, you include code which has nothing to do with this PR: you also add your timetracking code.
third: adding a checkbox is bad UI. the time format should be selectable in a dropdown.
i cannot merge it like this.

claudegel:
Ok I'll redo it without touching to the white space. I don't understand what you mean for 'selectable in a dropdown'. I just need standard time in the export module as some of my clients prefer standard time instead of decimal time. Could I place it in the preference ?

simonschaufi:
I mean you shoud not "tick" a checkbox for the different format but "select" the format that you want. that would be better.
About the UI: I'm thinking about an extension based setting but it's hard to integrate it in the current UI. So I think the best is to make a new tab in the settings for the export extension and put it there.

claudegel:
I don't understand what is the UI or the settings for export extention. My tick box was in the PDF export final setting (when you click on the PDF export button). I'll change the tick for a dropdown list. But instead of putting it in the pdf export setting maybe I can put it in the export top date-time format with a drop down there.

claudegel:
I've done another branch 'export_to_pdf' with all the corrections and drop down menu for standard and decimal time. Please close this one I'm gonna make a new PR as this one is not up to date and I'm not sure that just doing a commit is enough. Thank you
I've also removed all unrelated changes and put them in new PR

simonschaufi:
merged with #538

PR #436, opened 2014-12-12
Quick note

claudegel:
Add the ability to write quick note into timesheet while tracking. The quick note ability is selectable in preference. I've added an icon that appear in the active timesheet when the quick note is selected in preference.


simonschaufi:
@claudegel thank you for your contribution but there were some errors in the javascript in add_edit_timeSheetQuickNote.php

kevinpapst:
Is this bug fixed now? If not, please do a reverse merge.

simonschaufi:
I fixed the bug while merging

Issue #375, opened 2014-05-04
Remove setting "show_sensible_data"


kevinpapst:
There is no way this setting is required. 
Every sensible setting should ALWAYS be hidden!

PR #448, opened 2015-01-08
Japanese Language Localisation: (jp.php), English Language edits: (en.php).

TheWanderingCoder:
Japanese localisation files for Kimai Time Tracking.

Note there are some comments in this file.
Some entries need to be rearranged as per the comments due to the Japanese linguistical syntax. At most opportunities I tried to keep the same structure as the English syntax but this was not always possible. This leads to some of the other comments in which some lines are lo longer necessary and thus are left as blankspace. The remaining comments are merely asking what each section is I could not find them when using the GUI.

I did not have the time to translate the About message so that will probably happen next revision. Further, the syntax links for the time and date Admin Settings I have linked to their Japanese language pages on php.net. Finally, as I tried to match the English character width some statements are not as formal as their equivalent Microsoft-eske software suits. I aim to fix that next revision too.

I would have posted this on the forum but registrations are closed, thus I can not post.

Any problems, questions etc. Please let me know.

> **TheWanderingCoder**


kevinpapst:
We have massive spam problems, thats why the forum is closed for now. Will be reopened somewhen in the future. 
Thanks for your translation, very cool!

TheWanderingCoder:
Ah, that makes sense. Well when you eventually open it up again I shall give it a go. Until then I will stay on GH. By the way is it possible to get some clarification on some of the comments in my translation?

Sensible data: Not sure what this was.
By: I couldn't find this in the GUI but I may have just been tired.
Done By: Same as above.

Cheers and thanks for the quick reply.

core/language/jp.php
```diff
@@ -0,0 +1,468 @@
...
+"breiteDatumspalte" => "日付列の幅",
+"def_pctID"         => "既定プロジェクトID",
+"def_kndID"         => "既定客様ID",
+"def_evtID"         => "既定イベントID",
+"recommended"       => "おすすめ",
+"submit"            => "OK",
+"cancel"            => "キャンセル",
+"edit"              => "登録編集",
+"construction"      => "工事中...",
+"about"             => "について",			// This needs to go after Kimai. 　　ie. Kimaiについて
```
ServiusHack: Since 4a84c53665bc1cfebd17f93de8cb5992389ee15e "Kimai" is part of the translation string, so you can place it where appropriate.

core/language/jp.php
```diff
@@ -0,0 +1,468 @@
...
+"refundable_long"   => "返金　（社員に）",
+"refundable"        => "返金",
+"yes"               => "はい",
+"no"                => "いいえ",
+"comment_type"      => "コメントタイプ",
+"ctype0"            => "コメント",
+"ctype1"            => "おしらせ",
+"ctype2"            => "注意！",
+"hint"              => "ヒント:", 
+"usr_caution1"      => "",			// This line is not required in Japanese as it does not make grammatical sense to write anything before the subject <username>. The "If you change the user-name of" line has been added to the following line.
```
ServiusHack: `usr_caution1` and `usr_caution2` can be removed. They are not used anymore.

core/language/jp.php
```diff
@@ -0,0 +1,468 @@
...
+"designation"       => "指定",
+
+"cleared"           => "完了されてる",
+
+"atLeastOneGroup" => '少なくとも一つのグループを選択しなければならない',
+
+"noFading"          => "フェージング効果無しで",
+"user_list_hidden"  => "既定にユーザーリストを非表示する",
+
+'show_sensible_data' => 'デバッグエクステンションにセンシブルデータを表示する',			// What is sensible data?
```
ServiusHack: "Sensible" is a bad translation from German. Sensible data is data like the database password or the password salt. It is planned to always hide this data, see #375.

core/language/jp.php
```diff
@@ -0,0 +1,468 @@
...
+'pdf_headline'          => '労働時間と経費',
+'pdf_headline_only_times'   => '労働時間一覧',
+'pdf_headline_only_expenses'=> '経費一覧',
+'costs'                 => '費用',
+'expense'               => '経費',
+'summary'               => '概要',
+'full_list'             => '全リスト',
+'page'                  => 'ページ',
+'by'                    => 'by',			// What is this for?
+'done_by'               => 'done by',			// What is this for?
```
ServiusHack: This is shown in the list PDF export just before the username of a timesheet entry.

core/language/jp.php
```diff
@@ -0,0 +1,468 @@
...
+'finalamount'           => '合計',
+'pdf_headline'          => '労働時間と経費',
+'pdf_headline_only_times'   => '労働時間一覧',
+'pdf_headline_only_expenses'=> '経費一覧',
+'costs'                 => '費用',
+'expense'               => '経費',
+'summary'               => '概要',
+'full_list'             => '全リスト',
+'page'                  => 'ページ',
+'by'                    => 'by',			// What is this for?
```
ServiusHack: This is shown in the list PDF export just before the username of an expense entry. (Since 21ea4d89c76dba2dbdf35d3546e7e0b945cfc25a)

ServiusHack:
Your translation file is for Kimai 0.9.2 but we plan to release Kimai 0.9.3 where some translation keys were renamed. For example `knd` is now `customer` and `pct` is now `project`. Please compare your translation file to [language/en.php](https://github.com/kimai/kimai/blob/master/core/language/en.php) and make sure every key in the en.php file exists in your file.

PR #454, opened 2015-01-15
Update for en.php

TheWanderingCoder:
Added PHP end "?>" marker. Some spelling and grammar edits.

If there is something out of place please let me know. Also it is assumed this in American English. (Otherwise Cell would be Mobile etc.).

There is also a non consecutive use of 「 ' 」 and 「 " 」 if it means anything in particular (I am not sure exactly in which file each item is called so this is probably nothing).

There is room to add capital letters to certain field names for aesthetic purposes as well.

### DIFF
--- a/core/language/en.php
+++ b/core/language/en.php
@@ -54,7 +54,7 @@ return array(
     "edit" => "Edit Entry",
     "close" => "close",
     "construction" => "under construction...",
-    "about" => "About",
+    "about" => "About Kimai",
     "hours" => "hours",
     "mins" => "minutes",
     "secs" => "seconds",
@@ -232,7 +232,7 @@ return array(
     "testCustomer" => "Test-Customer",
     "testProject" => "Test-Project",
     "testActivity" => "testing",
-    "time_billable" => "Billable",
+    "time_billable" => "Billable Time",
     "billable" => "Billable",
     "budget" => "Budget",
     "effort" => "Effort",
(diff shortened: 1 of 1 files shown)


kevinpapst:
Hi!
First let me thank you for your effort in contributing to Kimai.
Two problems: 
- we can't merge this pull request due to conflicts 
- there is no reason why you ever want to have a closing php tag in pure PHP files

Please read why people do not include the closing PHP tag (if you don't trust me, check the reason why large frameworks do it as well).

TheWanderingCoder:
Hi Kevin,
1. What would the conflicts be? And is there some way to resolve them?
2. Ah I see. I wasn't sure because most of the other language files as well as the 0.9.2 en files had an ending `?>` in them. I am relatively new to PHP so I do still have quite a bit to learn.

kevinpapst:
The problem "We can’t automatically merge this pull request." might be (I am not 100% sure here) because your master shows "This branch is 1 commit ahead, 10 commits behind kimai:master".
You need to merge our original master first into your master.
Also there are some very good articles about providing "best pratice" pull requests: e.g. you don't want to sent in your master as pull request, but always a new branch.

I somehow can't get rid of all ?> ... ;-) but I will definitely not re-add them to the repo!

If you are new to Git and PHP, please don't be offended, but your pull request is nor mergeable. Easiest way to solve this: fix the metntioned problems and push them again => which will automatically update your pull request with the latest changes.

TheWanderingCoder:
Hmmmm. could have sworn I forked it about 10 mins before I put up the English fix. I will put the English File in the patch-1 branch, push that and see if that fixes it then.

I could try and kill some ?> for you but I have a feeling if there were some you needed I would probably catch those too. I shall remove them from the Japanese and English files that I have done though.

I am not offended at all! You gave me a probable reason for it not working and possible ways it will. I just have to make them fit the pattern.

_Update_
Added to the patch-1 branch and merged with the Japanese Language Pull request. Isn't appearing to say there is any issue.

kevinpapst:
Are you going to open a new pull request?

TheWanderingCoder:
I included it in the other pull request I had and changed the name to reflect this. I can split them up if you would like.

PR #476, opened 2015-03-23
update permissions error text during installation

kralo:
explain that not only the /includes/autoconf.php needs writing permissions, but you also want the whole /includes directory


ServiusHack:
The `includes/autoconf.php` file will not exist for new installations, so the installer needs the permission to create this file. I added the file some time ago but we had to remove it again, see #277 and #349. A redirect to the installer will only happen if this file is missing, see `includes/basics.php` line 47.

It's not that great to require the whole directory to be writable. A pull request which frees us from this requirement will be very much appreciated.

PR #487, opened 2015-05-24
Export bugfixes

simonschaufi:
FIX: recalculation of table header
Adjust column widths so that it aligns properly with the header
Optimize jQuery selectors
Fix: charset of rendered printable table


simonschaufi:
ups, combined 2 pull requests here.

PR #513, opened 2015-09-16
[FEATURE] added an button to the display to set the current date as timespan

plessmann:
Added a new button/icon for quick select of the current day only

2015-09-16 20_00_09-lars - kimai


kevinpapst:
Hi @plessmann, can you tell us where this icon is coming from (license)?

PR #518, opened 2015-11-08
Russian language support. Not all translated, but good start.

21h:
peace


simonschaufi:
thank you very much for your contribution!

Issue #526, opened 2015-12-22
Excel export: Field formatting & sum formulas


lentschi:
Currently all the fields in the excel export - including decimal numbers - appear to be formatted as text, which makes summing up hours difficult. (At least in LibreOffice Calc the field format has to be changed manually and then the contents have to be replaced with real numbers.)

So I suggest the following changes:

1. Add formatting to the fields according to their type (e.g. a date should have the Excel format `YYYY/mm/dd` or whatever matches config's `date_format_1`, wages the Excel format `0.00` prepended by config's `currency_sign` and so on...) This would make the decimal delimiter field in the xls export floater obsolete, as that would be locale dependant.
1. Add sum formulas for columns, where it makes sense
1. General formatting, that reduces the effort to create a version that can be printed or exported on a single PDF page:
    * emphasize headers and sums using borders and bold font
    * thicker lines between two different days
    * landscape page format (When selecting only a few columns for export, there is a chance all of them might fit on a single page)

I have already implemented those requests in [my fork](https://github.com/lentschi/kimai), since I needed a time list format like this for a customer of mine. However, if you think, that this is something the majority of your users would want, but you want some specifics changed/configurable, go ahead and tell me - I still might find some spare time. ;-)

PR #529, opened 2015-12-24
[BUGFIX] csv export: description (resolves #525)

### LINKED ISSUE #525: CSV export: Description is missing
The description field's data is missing from the CSV export (no matter if the column has been disabled or not in the export tab)

lentschi:
added `description`-field to CSV export (was missing even
when column the column was enabled)


lentschi:
Whoops - that commit should be separated (inlcudes part of #526)! Closing the PR and creating a new one....

PR #532, opened 2015-12-25
missing translation in en.php and fr.php

claudegel:
small changes in two translations

### DIFF
--- a/core/language/en.php
+++ b/core/language/en.php
@@ -475,6 +475,7 @@ Choose another table prefix or delete existing tables with this prefix in order 
     'ext_invoice' => array(
         "invoiceTemplate" => "Invoice template:",
         "invoiceTitle" => "Create invoice",
+        "invoiceProject" => "Projects:",
         "invoiceTimePeriod" => "Time Period:",
         "invoiceOptionShort" => "Short Invoice",
         "invoiceOptionRound" => "Round time",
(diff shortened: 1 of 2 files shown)


core/language/en.php
```diff
@@ -475,6 +475,7 @@
     'ext_invoice' => array(
         "invoiceTemplate" => "Invoice template:",
         "invoiceTitle" => "Create invoice",
+        "invoiceProject" => "Projects:",
```
simonschaufi, line 478: this is nowhere used in the code and can actually be removed completely from all lang files. I'll do the cleanup in another commit.

claudegel, line 478: Le Sun, 27 Dec 2015 10:32:51 -0800,
Simon Schaufelberger notifications@github.com a écrit :

> > @@ -475,6 +475,7 @@
> >      'ext_invoice' => array(
> >          "invoiceTemplate" => "Invoice template:",
> >          "invoiceTitle" => "Create invoice",
> > -        "invoiceProject" => "Projects:",
> 
> this is nowhere used in the code and can actually be removed
> completely from all lang files
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/kimai/kimai/pull/532/files#r48457118

Is it all the 'ext_invoice' => array( that is not used anymore or just
these three line

> > ```
> >      "invoiceTemplate" => "Invoice template:",
> >      "invoiceTitle" => "Create invoice",
> > ```
> > -        "invoiceProject" => "Projects:",

I can remove all them in all language file if you want

## 

Claude Gélinas agr.
Phyto Ressources Inc

simonschaufi, line 478: just the invoiceProject but I think it was an error in the extension since the label was not correct there. I have changed it already so that the label is used.

PR #534, opened 2015-12-29
 [FEATURE] excel export: formatting (resolves #526)

### LINKED ISSUE #526: Excel export: Field formatting & sum formulas

lentschi:
s. #526:
I added formatting and sum formulas to the excel export.
This is a complete rewrite of the excel export and uses `PHPExcel` (currently included as submodule) to generate xlsx files.
You will need to run `git submodule update --init`, should you want to try it out.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "core/libraries/PHPExcel"]
+	path = core/libraries/PHPExcel
+	url = https://github.com/PHPOffice/PHPExcel.git
--- a/core/extensions/ki_export/processor.php
+++ b/core/extensions/ki_export/processor.php
@@ -276,24 +276,16 @@ switch ($axAction) {
     case 'export_xls':
 
         $database->user_set_preferences(array(
-          'decimal_separator' => $_REQUEST['decimal_separator'],
           'reverse_order' => isset($_REQUEST['reverse_order'])?1:0),
           'ki_export.xls.');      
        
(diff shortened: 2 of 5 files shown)


simonschaufi:
I would prefer if the library would be included via composer but then it is not in the desired folder (core/libraries). With composer you can make sure you always get a working version. Let's figure out how we can place it there without too much trouble.

lentschi:
sure can do :)

PR #535, opened 2015-12-30
 [FEATURE] excel export: formatting (resolves #526)

### LINKED ISSUE #526: Excel export: Field formatting & sum formulas

lentschi:
added formatting and sum formulas to the excel export
uses `PHPExcel` (included using composer) to generate xlsx files

To try it out, you will need to run `composer install`.

@simonschaufi 

> I would prefer if the library would be included via composer but then it is not in the desired folder (core/libraries). With composer you can make sure you always get a working version. Let's figure out how we can place it there without too much trouble.

I just had to add `vendor-dir: "core/libraries"`.

It would be pretty tricky, if you'd want to have `phpunit` installed to `vendor/`, but everything else to core/libraries/. (There is https://github.com/mnsami/composer-custom-directory-installer, but that didn't work for me...)

I don't think this is necessary though. I added `bin-dir: "bin"`, so now you can run phpunit by entering `bin/phpunit`. People who just want to use kimai for production will run `composer install --no-dev` anyway.

In the long run I'd suggest making all the packages under `core/libraries` composer packages. (Than you could change the `.gitignore` file there to ignore everything but itself.) If not possible, you should maybe add an extra subpath and change composers `vendor-dir` to `core/libraries/composer`

By the way, I added composer.lock to git as it [should be in there](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "time": "2013-08-03",
     "version": "0.9.3-beta1",
     "config": {
-        "vendor-dir": "core/libraries",
+        "vendor-dir": "core/libraries/composer",
         "bin-dir": "bin"
     },
     "authors": [
@@ -36,22 +36,5 @@
     },
     "include-path": [
         "core/libraries"
-    ],
-    "repositories": [{
-	    "type": "package",
-	    "package": {
-	        "name": "PHPOffice/PHPExcel",
-	        "version": "1.8.0",
-	        "source": {
-	            "url": "https://github.com/PHPOffice/PHPExcel.git",
-	            "type": "git",
-	            "reference": "1.8.0"
-	        },
-	        "autoload": {
-	            "psr-0": {
-	                "PHPExcel": "Classes/"
-	            }
-	        }
-	    }
-	}]
+    ]
 }
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,10 @@
     "minimum-stability": "dev",
     "time": "2013-08-03",
     "version": "0.9.3-beta1",
+    "config": {
+        "vendor-dir": "core/libraries",
+        "bin-dir": "bin"
+    },
     "authors": [
         {
             "name": "Severin Leonhardt",
@@ -19,7 +23,8 @@
         }
     ],
     "require": {
-        "php" : ">=5.3.0"
+        "php" : ">=5.3.0",
+        "PHPOffice/PHPExcel": "1.8.*"
     },
     "require-dev" : {
         "phpunit/phpunit" : "3.7.*"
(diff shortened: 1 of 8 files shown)


simonschaufi:
thank you for your investigation. btw you don't need to open another PR always. just go to your branch, make your changes, commit them and push again and the PR will be updated automatically.

composer.json
```diff
@@ -31,5 +36,22 @@
...
+	            "type": "git",
+	            "reference": "1.8.0"
+	        },
+	        "autoload": {
+	            "psr-0": {
+	                "PHPExcel": "Classes/"
+	            }
+	        }
+	    }
+	}]
```
simonschaufi: since this library is also on https://packagist.org/packages/phpoffice/phpexcel, i don't think we need this, or?

lentschi:
> btw you don't need to open another PR always. just go to your branch, make your changes, commit them and push again and the PR will be updated automatically

Thanks, I know ;-) I just didn't wan't you to have thousands of commits for a single feature...

But now I pushed two new commits on the same branch:
- updated composer:
  - You were right: There is no need for the extra package source - thanks for checking!
  - I moved composer's `vendor-dir` one level down to `core/libraries/composer` (as I was pondering above), because else developers might accidentally commit packages added via composer. But of course it's up to you how you want to structure this...
- Added a small bugfix for the excel export itself

simonschaufi:
before I can merge this, i need to talk with the team since the demo website installer also needs to be adjusted to execute a composer install after installation.

lentschi:
Alight, thanks for the info!

simonschaufi:
merged in develop with https://github.com/kimai/kimai/commit/57ec3a147b84d7485b9e517897e809a172e405be

PR #537, opened 2016-01-03
Improve ki.export disabled column appearance

claudegel:
A small change to put a line through and a font colour change when we click on column title to disable them. Just a column colour change is not enough to understand that this column is disabled.

### DIFF
--- a/core/extensions/ki_export/css/tables.css.php
+++ b/core/extensions/ki_export/css/tables.css.php
@@ -21,11 +21,6 @@ div.ki_export table a:hover {
     border-bottom:none;
 }
 
-div.ki_export table a.disabled {
-    text-decoration:line-through;
-    color: #FFC7BB;
-}
-
 div.ki_export table thead {
     height:25px;
     text-align:left;
@@ -253,6 +248,10 @@ div.ki_export table tr.even td.disabled {
   background-color:#FFE1DB !important;
 }
 
+div.ki_export table a.disabled{
+    text-decoration:line-through;
+    color: #FFC7BB;
+}
 
 div.ki_export table td.user
 {
--- a/core/extensions/ki_export/css/tables.css.php
+++ b/core/extensions/ki_export/css/tables.css.php
@@ -21,6 +21,11 @@ div.ki_export table a:hover {
     border-bottom:none;
 }
 
+div.ki_export table a.disabled {
+    text-decoration:line-through;
+    color: #FFC7BB;
+}
+
 div.ki_export table thead {
     height:25px;
     text-align:left;


core/extensions/ki_export/css/tables.css.php
```diff
@@ -21,6 +21,11 @@
     border-bottom:none;
 }
 
+div.ki_export table a.disabled {
```
simonschaufi: `a.disabled` doesn't exist. only `td.disabled a`

claudegel:
OK but this way it is doing what we want
export-screen

td.disabled affect only the disabled column color. But in the title, because it is clickable there should be a <a... so a.disabled should exist when you click on a column title to disable it.

I've changed the code for 

``` css
div.ki_export table td.disabled a{
    text-decoration:line-through;
    color: #FFC7BB;
}
```

and it work too 
only the right column do not get the line through when disabled

a.disabled give that behaviour:
a disabled
only some column get the line through decoration
td.disabled a give this:
td disabled_a
all column except comment, location and tracking number do not have the line through when disabled
a.disabled look better as it affect the columns that we want to disable so I've move it close to the other td.disabled in the css

simonschaufi:
when i checkout your code, only the tracking number column has a.disabled. when i disable the customer column for example, my html code looks like this:

``` html
<td class="customer disabled" style="width: 149px;">
    <a onclick="export_toggle_column('customer');">Customer</a>
</td>
```

are you on the latest master branch?

This can be reproduced on http://demo.kimai.org/core/kimai.php

claudegel:
I'm on the latest master I think f43bc48
if I disable the «Rate» and «Total» column my html code is like this

``` html
<td class="rate disabled" style="width: 50px;">
    <a class="rate disabled" onclick="export_toggle_column('rate');">Rate</a>
</td>
<td class="wage disabled" style="width: 50px;">
<a class="wage disabled" onclick="export_toggle_column('wage');">Total</a>
```

It look like only the a.disabled column get the line through

for customer I get the html like you get

``` html
<td class="customer disabled" style="width: 91px;">
    <a onclick="export_toggle_column('customer');">Customer</a>
</td>
```

so a.disabled will only affect the following column:
Rate, Total, Comment, Location and tracking number which is fine. If we need to be able to disable all of them, then I need to check in the code to find out why each column do not get de disabled class for the <a

simonschaufi:
i have just updated your comment to use proper styling ;)
PS: use 3x ` to make code blocks. edit your comment to see how :D

claudegel:
Better like this :)

claudegel:
By using a.disabled it is conform to what we have in ki_export/templates/scripts/main.php.
Only the four columns: Rate, Total, Comment, Location and tracking number can have a.disabled class, other one no. this is the behaviour I see when using a.disabled instead of td.disabled a

claudegel:
Sorry columns time and dec_time can are also be disabled since last merge

claudegel:
I see that in the export_pdf2.php, almost all column are used and can be disabled. So, if we change all columns definition in the <tr> line in ki_export/templates/scripts/main.php to add the class in the <a section like this line

```
<td class="time"><a class="time <?php if (isset($this->disabled_columns['time'])):?> disabled 
<?php endif; ?>" onClick="export_toggle_column('time');"><?php echo $this->kga['lang']['time']?></a></td>
```

the a.disabled will work with each column, better then the td.disabled a which is not working for the last three columns for some reason.

claudegel:
dont know why but my last comment do not show properly

simonschaufi:
you need to write backticks, that frensh "thing" on the à ;)

claudegel:
Ha now it work thank's

claudegel:
I've changed all column in export_pdf2.php so they can be disabled/enabled via 'a.disabled' in the css. This also improve column selection for the export to pdf2 so we can play with almost any column. I also put a line through for all line of the disabled column so it is more evident that the column is not available for export. Should I proceed and add theses last changes to this pull request or you won't consider this pull request at all. Let me know

PR #538, opened 2016-01-03
Export to pdf in decimal or standard time

claudegel:
This PR replace the export_pdf PR with the chages we have discussed. I've also remove the unrelated code that will show up in other specific PR.


simonschaufi:
thanx claude

PR #540, opened 2016-01-03
wrong path for css and logo file

claudegel:
in  core/extensions/ki_invoice/invoices/05_HTML_Invoice/index.html 
there was an error in the path of two file, style.css and logo.png


simonschaufi:
works for me, see demo website. why is this path incorrect?

kevinpapst:
Never had a problem either

claudegel:
Sorry, this was reported by my ZendStudio as a wrong path. But if it is not broken better not fix it :)
Just ignore it

PR #542, opened 2016-01-04
[BUGFIX] Fix wrong id error when using "Open entry after the recording was stopped" option (resolves #541)

### LINKED ISSUE #541: Bug regarding the "Open entry after the recording was stopped" option
When enabling the "Open entry after the recording was stopped" option in the user preferences, when stopping the timer, a wrong popup appears, because the does not get the correct id (id is empty actually).

Sending a pull request with fix in a moment.

smoove:
Fix wrong id error when using "Open entry after the recording was stopped" option (resolves #541)

### DIFF
--- a/core/js/main.js
+++ b/core/js/main.js
@@ -365,9 +365,10 @@ function stopRecord() {
         function(response){
               ts_ext_reload();
               document.title = default_title;
-              if (openAfterRecorded)
+              if (openAfterRecorded) {
                 var data = jQuery.parseJSON(response);
                 editRecord(data['id']);
+              }
         }
     );
 }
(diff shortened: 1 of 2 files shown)


core/js/main.js
```diff
@@ -362,11 +362,12 @@ function stopRecord() {
     $("#timeSheetTable>table>tbody>tr#timeSheetEntry"+currentRecording+">td").css( "color", "#FFF" );
     show_selectors();
     $.post("processor.php", { axAction: "stopRecord", axValue: 0, id: currentRecording},
-        function(){
+        function(response){
               ts_ext_reload();
               document.title = default_title;
               if (openAfterRecorded)
-                editRecord(currentRecording);
+                var data = jQuery.parseJSON(response);
```
simonschaufi, line 369: i think you forgot putting the block in { ... } here, or?

smoove:
Yeah, sorry i'm not used to that coding style. Fixed now.

PR #545, opened 2016-01-05
Added function "record duration only"

cljk:
We have the need to record just the time spent on a specific activity per day and user - not the beginning and the end... just the duration.
- fixed a bug on changing duration field - it should change the end time not the beginning time as written in code comment
- added a new checkbox in configuration "record duration only" (config key: "record_durationOnly")
- when option is set:
  - hide "from/to" columns in display and export
  - on creating new entry: set "from/to" to "00:00" instead of current time - usecase is changing the "duration"-field or "end"-field
- added the timepicker to the duration-field to quickly select hours/minutes

This morning I merged the current kimai master so that my branch is mergable

In a next step (later) I want to hide "from" and "to" fields in new entry form when option is set.


cljk:
"from" and "to" fields in new entry form are hidden now when option "record duration only" is set

cljk:
Any news to that pull request? For me it's function complete...

if you don't want to merge just tell. Then I know that we have to keep our own branch...

simonschaufi:
@kevinpapst I don't want to decide this alone.

kevinpapst:
Is that compatible with the other extensions (e.g. invoice)? 
I have not tested it yet, but it sound like a major change to the core functionality of Kimai.

@skilleven whats your opinion?

cljk:
It's mostly a cosmetic change - no data structures are changed, just the fields are hidden. So I think all other functions shouldn't be affected.
On adding a new entry "startDate" is always "00:00" and "endDate" is equal to duration. 
It's all for data security reasons we have to fullfill in our company (@ germany)

I tested export, pdf print and so on. If testing the invoice plugin is an issue I'll do it in ~2 wks. (on vacations right now ;-) )

kevinpapst:
Hm, okay. We are from Germany as well ... but I am self-employed, so I don't have that data security concern ;-) All the companies I worked for in the past wanted to know when their employees come and leave, but nevertheless, I see the use-case for that feature.
I need some time to test that feature, but its likely that we will integrate it into the core.

One question that comes to my mind: should the setting to dis-/enable that feature be globally or is that something, that a single user might want to change on its own?

cljk:
We have a different system for presence-times, holidays etc - we use kimai to track hours spent on specific projects and activities. And there comes our "Betriebsrat" and says: please dont track the beginning/end timestamp - just how many hours are used for X and Y. It's also better for "rounding" ;-)

Enable/disable global vs user specific: didn't think about that - had no need to specify per user (for us) - so made it globally.

matecode:
Any news on this topic? We would love to have this feature.

cljk:
@matecode Since my code was working but not merged .... and it´s usable for us and is in productive use since ~2 yrs I freezed my kimai installation and won´t update and am not going to upgrade my modifications the next months. Feel free to do that ... 

If you´re okay with using an old version of kimai you can use my fork we´re using...

https://github.com/cljk/kimai

mawa85:
I also have demand for such a feature. It would be great if this could be merged

matecode:
@kevinpapst Are there any plans for this feature? Without such a feature it wont be possible for most german companies to use Kimai at all. We are a small company, and also dont track time-windows but only durations.

simonschaufi:
What are your requirements for this feature? Can the time be split in chunks or does it need to be like 3h + 4h = 7h?

mawa85:
thanks for your reply.

for me, chunks are fine. 
I just need an option to "hide"  start and stop times in the input mask and timesheets

matecode:
@simonschaufi Thanks for your answer. 

If I understand you right, chunks are totally fine. I think there arent any needs to change the kimai internal behavior.

So the correct needs are what this pull request originally did:

- Hide all FROM/TO fields in timesheet, add entry, export ...
- Timepicker for Duration-Field: hours:minutes ... so you d'ont have to manually type text in duration field

Combined you can say to make data-privacy version, there could be a "Duration Only" configuration option

Carm66:
It would be really great if there could be a "Duration Only" option.
In my company we only need to record the time spent in our different tasks and it's quite painful to artificially add start and end dates :/

cljk:
If there are at least four (?) persons in need of this function... it should be possible that one of them compares my modifications and adapts them to the current code base - eleminating the merge conflicts?!

kevinpapst:

PR #550, opened 2016-01-20
rate calculation for remote API

Viras-:
Entries added via the JSON API do not have the rate set based on user / project / activity. This small change fixes this issue.


kevinpapst:
Thanks!

Viras-:
You are welcome. Maybe can you consider adding this as a hotfix to the current release version? Otherwise all entries added with the app will have a rate / wage of 0.0!

PR #558, opened 2016-02-23
Refactored directories

kevinpapst:
This finally cleans up the old and "wrong" directory structure.
It will allow us to move forward with composer/dependency management and make distribution a lot easier.


simonschaufi:
:heart: :heart_eyes: Thank you sooo much!!!111elf
I'm fine with a merge or do you want to add more commits?

kevinpapst:
Thanks :) I'd like that to be merged, but as it has conflicts , I am going to resolve them later and comment again.

simonschaufi:
I changed just one file yesterday. Didn't think that this would cause a merge conflict but it did

kevinpapst:
Resolved conflict, @simonschaufi if you want you can merge.

simonschaufi:
Now it would be nice if core/kimai.php could also be replaced with index.php

I actually like the idea how TYPO3 is doing it. They have the sources in the root directory and then a web folder with just a .htaccess and index.php in it. The advantage is that the sources are not directly accessible.

I our case the libraries folder for example is such a folder which doesn't need to be accessed via URL.

kevinpapst:
Definitely. There is so much room for improvements ...
As you said, then moving all javascripts and images to the public folder, including the index.php and the json/soap server as well.
I'll see what I can do.

simonschaufi:
I would also like to merge #535 and remove some libraries out of our repo and include them again via composer. then you need to adjust the demo server again. when are you available for that?

kevinpapst:
Please wait until tonight - I Have a huge refactoring coming down the way, including changes to composer and autoloading: https://github.com/kevinpapst/kimai/tree/cleanup
Any commit would cause major merge trouble right now, so please don't do it yet!!!!

PR #559, opened 2016-02-28
composer Autoloading / folder and file cleanup and more

kevinpapst:
Major refactoring:
- using composer for autoloading instead of Zend1 Autoloader
- using composer to load Zend Framework 1 (thats why 6000 files have changed)
- committing Zend Framework 1 nevertheless. Until we have changed the build.xml for releasing Kimai, I want to keep everything for a release inside the repo (this is going to be changed as well, but not right now)
- adding a global skin variable, allowing us to use skins in login/error screens as well 
- adding a billable config flag in autoconf.php and a fallback in vars.php, allowing to remove the config/ folder and the Config.php class, which was used only for that file
- moving all CSS files from the css/ folder to skins - getting rid of that folder
- moved some of the images from grfx/ to skins as well, the remaining ones are used by updater/installer (don't know what to do with them right now)
- removed shellscripts folder (added a new repo for that https://github.com/kimai/scripts , end user don't need these scripts)
- moved all classes from includes/classes/ to library/Kimai/ where they belong
- removed multiple unused files


kevinpapst:
Please don't merge right now. 
I have to test installer/updater and remote classes a bit more, before including that in the master branch.
Just wanted to share that huge refactoring, so everyone knows where the journey is going to.

kevinpapst:
Okay, updater/installer/json remote worked just fine. 
Now only travis must accept the changes ;-)

kevinpapst:
@simonschaufi Any thoughts on that one or are these changes just to huge to review?

simonschaufi:
there are a few questions i have:
- why did you move zend in a sub folder? isn't it possible to autoload it just like that?
- whats this billable config flag used for? i never understood the usecase of the ini file
- any code style you want to use in the future? I have seen you did some code formattings. I actually want to use PSR standards but some formattings you made don't make sense in that case like aligning assignments.

Apart form that, go ahead, I don't care about the size of a PR. You have explained what will be merged. That's fine for me.

kevinpapst:
The billable config is used in the time-entry floater. There is a drop-down in the "Budget" tab called "Verrechenbar". I have no clue what its for and I don't know who added it in the first place. 
Seems to be a text column that can be used for controlling expenses/invoices??? 
I just moved it to autoconf, which I believe is the proper place, as it won't be overwritten by an upgrade.

I didn't move zend to the subfolder, composer did! It seems to be the way the ZF1 will be installed when using composer. Yes, the folder structure was cleaner before, but hey, if thats the "price we have to pay" to manage our dependencies with composer, I am willing to pay it ;-)

kevinpapst:
Its that configurable by composer? I don't care, I am used to calling it through the sub-directory. Thats the way I always used it (in all projects I used composer so far). But if you want to change it later, I don't mind if you do so.

simonschaufi:
about billable: ok, need to dig deeper in the code to understand how this is used.
about zf: ah ok. maybe because of the namespaces. The typical format is /vendor/library/root_folder_of_code... so in our case if we namespace kimai, it would be like /KimaiTeam/kimai/...

did you see this? https://github.com/kevinpapst/kimai/commit/32564e138f17cf102345447bcee86e6ee43b98b3#commitcomment-16376201

I'm fine with a merge now.

kevinpapst:
Yes I know, I guess its because ZF1 was created long before composer rise up. 
They distribute it that way .. unfortunately.
Did you see my comment: https://github.com/kimai/kimai/pull/559#issuecomment-189872315
We can change the bin dir, but lets do that later. The merge is already to big ;-)

I will merge it now, then we can go on with more improvements!

PR #563, opened 2016-02-28
Updater cleanup

kevinpapst:
This pull request refactores the updater.php:
- create a new updater/ directory containing all related files
- moved db_restore.php htere
- moved images from grfx/ to updater/ and deleted grfx/
- created header/footer and reusable error templates, to separate HTML from PHP
- removed updates before Kimai 0.8 as they were broken and needed PDO support. Added an error message and download link to older Kimai version which has update support for these very old versions.


kevinpapst:
Enough refactoring for now ;-) after that Pull request I am going to concentrate more on functionality and bug fixing. Before merging I am going to squash all commits...

kevinpapst:
Created a new Pull request #564 with squashed commits

PR #567, opened 2016-02-28
Remove support for multiple database

### LINKED ISSUE #560: Remove support for mutiple database
Kimai has had the feature to support usage of different databases, through Request/Cookie settings.
I never used the feature, never heard any user asking for such a feature or complaining that it doesn't work nor did I ever find any documentation about how to use it.
This feature is still kind of active in https://github.com/kimai/kimai/blob/master/includes/basics.php#L68 and following lines.

I would recommend that we delete this feature, which I believe does not work anymore.

Thoughts on that @kimai/core-team ?

kevinpapst:
This pull request removes some very old code, which could be used to switch between different databases.

This feature was never used and imposed a security risk, therefor we deleted it.

Closes #560

### DIFF
--- a/includes/basics.php
+++ b/includes/basics.php
@@ -26,88 +26,48 @@
 defined('WEBROOT') || define('WEBROOT', dirname(dirname(__FILE__)) . DIRECTORY_SEPARATOR);
 defined('APPLICATION_PATH') || define('APPLICATION_PATH', realpath(dirname(__FILE__) . '/../'));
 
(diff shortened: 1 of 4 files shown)


simonschaufi:
@kevinpapst took me some time to merge this PR as you also changed some code formattings. since I'm doing this with scrutinizer, i had quite some trouble

kevinpapst:
If we are really going to work on our code style, we have to discuss the format we want to use. Thats a team decision. And thanks!

PR #569, opened 2016-02-28
Minor translation changes

kevinpapst:
- changed about translation (only in de)
- changed footer links in login screen
- fixed duplicate "Kimai" in link to about screen

@simonschaufi added your name to the about screen :+1:


simonschaufi:
@kevinpapst thanx! btw, its the 1.0.0 milestone that i always use.

kevinpapst:
Hm, and I used 1.0 all the time. We have to move those lists together probably and delete one of them,

simonschaufi:
yes, I will rename the 1.0.0 later to 1.0

PR #578, opened 2016-03-01
Move ultimatemysql into correct library folder

simonschaufi:
phpclasses also provide a way to include their classes via composer. Unfortunately only with username and password. Thats why i hardcode them into the library folder without putting the reference in the composer.json


kevinpapst:
I miss a commit for the composer autoload files. 
With my changes, the class is not any longer required in our code.
I wonder how that works ... or did you commit while you had composer on the ignore list?

Can you please call `composer dump-autoload` and check if there are changed files inside libraries/composer/?

kevinpapst:
Oh and btw: did you install it via composer? Maybe we should add a new page in the documentation about composer and that class.

simonschaufi:
yes i was a bit afraid of commiting the composer stuff as well but now i see that it is required to work.
as i said, i would like to add it to the root composer.json file as well but phpclasses require authentication for whatever reason... :confused:

kevinpapst:
Yes, the authentication was a reason why we never updated it before ;-)
I would not add it to composer.json yet, because everybody would need to have an account at phpclasses to update any package via composer. 
But please commit the generated autoload files (the path change must be reflected there as well, to be able to autoload it correctly).

PR #587, opened 2016-03-01
Fix open_basedir problems

kevinpapst:
In environments utilizing the open_basedir setting from PHP, there occured error logs, because ZF1 tried to open files within the complete include_path - which might have path settings set that were outside the open_basedir limitations.
It does not affect Kimai itself, but it spams the logfiles.

This code was previously available, but was accidentaly deleted with the autoloading feature of composer, so this is only a revert commit.


kevinpapst:
Merci!

PR #588, opened 2016-03-02
Fix E_NOTICE Undefined offset: 3

simonschaufi:


kevinpapst:
Whoa, that file needs a code style update later ;-)

PR #590, opened 2016-03-02
Credits

kevinpapst:
This PR works on improving our options to change the credits screen in the future.
Before the Credits screen was one large translation entry per language, which made it (kind of) impossible for us to control the content.

Now its divided into several sections:
- intro => developer names - not translatable, comes from the template
- links => developer names - not translatable, comes from the template
- license => translatable, link injected via sprintf
- free translation => default text from the current translations, including links to the most important pages like forum, issue tracker, donation page ... links injected via sprintf
- further thanks => label translatable, list of names is fix within template
- links to projects => label translatable, list of projects and links is fix within template

There were several outdated links in the translations (still linking to sourceforge for example) which can now be changed by us in one single place.

I also updated the footer in the welcome/login screen. I removed the "powered by" line, which btw. included links to wrong projects (like smarty which we don't use any longer). I also shortened the GPL link, as it is not truly relevant to Kimai at this place


language/cz.php
```diff
@@ -283,48 +283,18 @@
...
-        <a href='http://sourceforge.net/projects/kimai/' target='_blank'>Sourceforge</a> |
-        <a href='http://www.kimai.org/en/download.html' target='_blank'>Download</a>
-        
-        <br /><br />
-
-        If you do any alternations we would appreciate a posting on the <a href="http://forum.kimai.org" title="Kimai Forum">forum</a>.
-        You'll find a special board "<a href="http://forum.kimai.org/index.php?board=10.0" title="Extensions / Friendly Hacks">Friendly Hacks</a>" there.
-        Latest Developer Version can be downloaded <a href="https://github.com/kimai/kimai/archive/master.zip">here</a> (download for testing purposes only!).
-        
+    'credits' => <<<'EOD'
```
simonschaufi, line 286: new syntax: [nowdoc](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc)

kevinpapst:
Yes, because heredoc cannot be used with sprintf.

PR #591, opened 2016-03-02
Fix #546

### LINKED ISSUE #546: language mixup with check for update
Kimai have been installed with french language. Then switch to english in preference and admin panel. If I log out then the login screen is still in french. when I log in again and go in admin panel, everything is in english  as it should be but if I click on "look for update" I get a french answer.
if I go to preference and switch to french, then in admin panel all is in french and when I check for update I get a french answer. Switching for english again I still have a french answer for the look for update.

kevinpapst:
This PR does two things:

1) reformatted code in ki_adminpanel/processor.php and added brackets for every multi line statement (if, else ...)
2) fixes #546 

There was a problem with writing the autoconf.php after updating extended settings. The used language for saving was the one from $kga array not from the request. So it would be updated, but the written language was still the old one. After saving again, the new language would then be written to the autoconf.php
So you always had the old language on all screens that relied on the autoconf.php instead of the DB config setting (like the login screen).

### DIFF
--- a/extensions/ki_adminpanel/processor.php
+++ b/extensions/ki_adminpanel/processor.php
@@ -16,195 +16,217 @@
  * You should have received a copy of the GNU General Public License
  * along with Kimai; If not, see <http://www.gnu.org/licenses/>.
  */
-// ================
-// = AP PROCESSOR =
-// ================
-// insert KSPI
+
 $isCoreProcessor = 0;
 $dir_templates = "templates/";
-require ("../../includes/kspi.php");
+require "../../includes/kspi.php";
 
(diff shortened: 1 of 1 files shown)


kevinpapst:
The GitHub diff is not very good, use your IDE if you are interested in more detail. 
The real change is here: 
https://github.com/kimai/kimai/pull/591/files#diff-f313b5440dfa3185cd8c9d7c8faadd62R597

All the rest were manual style changes (please do not check every line, I just added brackets and fixed indentation). As discussed, I reformatted the file while working on it!

PR #592, opened 2016-03-03
DB access refactoring (fixes #566, #501)

### LINKED ISSUE #501: Incomplete delete - users
Hello again,

I'm looking at the admin->membership roles, and I see a users count of 20 on the admin role... but there are only 5 users...

So, Mysql->user_delete() does not include the following related tables
  groups_users
  preferences
  rates
  timesheet
I will leave timesheet alone because of possible negative consequences, but the other tables should be cleaned of the deleted user, otherwise we are leaving dead data.

So here's my version of function user_delete():
file: libraries/Kimai/Database/Mysql.php
```php
    public function user_delete($userID, $moveToTrash = false)
    {
        $userID = MySQL::SQLValue($userID, MySQL::SQLVALUE_NUMBER);
        if ($moveToTrash) {
            $values['trash']  = 1;
            $filter['userID'] = $userID;
            $table            = $this->kga['server_prefix'] . "users";

            $query = MySQL::BuildSQLUpdate($table, $values, $filter);

            return $this->conn->Query($query);
        }

        $query  = "DELETE FROM " . $this->kga['server_prefix'] . "groups_users WHERE userID = " . $userID;
        $result = $this->conn->Query($query);

        if ($result === false) {
            $this->logLastError('groups_user_delete');

            return false;
        }

        $query  = "DELETE FROM " . $this->kga['server_prefix'] . "preferences WHERE userID = " . $userID;
        $result = $this->conn->Query($query);

        if ($result === false) {
            $this->logLastError('preferences_delete');

            return false;
        }

        $query  = "DELETE FROM " . $this->kga['server_prefix'] . "rates WHERE userID = " . $userID;
        $result = $this->conn->Query($query);

        if ($result === false) {
            $this->logLastError('rates_delete');

            return false;
        }

        $query  = "DELETE FROM " . $this->kga['server_prefix'] . "users WHERE userID = " . $userID;
        $result = $this->conn->Query($query);

        if ($result === false) {
            $this->logLastError('user_delete');

            return false;
        }

        return true;
    }
```
Ideally, it's probable that a check should be done on the other table, that depends on the user entry, before performing a delete (that's probably why the trash was put there in the first place).

correct me if you see fit. :)

Anyway, that's it for now.

c.

### LINKED ISSUE #566: Remove unused variables
The following variables are still spread within the codebase, even though they are not used:

- server_type
- server_conn
- $kga['server_type']
- write_config_file() => parameter 5 $db_layer and parameter 6 $db_type
- $db_layer    = $_REQUEST['db_layer'];
- $db_type     = $_REQUEST['db_type'];

These are left overs from the time when we had multiple DB layers.
They are currently kept in autoconf.php and used for example in the constructor of the Mysql class.
We should eliminate all usages of these variables once and for all.

kevinpapst:
- removed Kimai_Database_Abstract, as we do not have multiple DB layers (anymore)
- removed all occurences of server_type and server_connection as we do not support different database types
- adjusted write_config_file to not write server_type and server_conn anymore (shrinks autoconf.php)
- code style/reformatting/brackets in our Mysql class
- added autoconnect feature in our MySql class

Fixes #566
Fixes #501

PLEASE DO NOT MERGE YET! I will ping you if I am ready.


kevinpapst:
Replaced by #597, due to broken commit history.

Issue #469, opened 2015-02-25
How to fully delete, or reactivate deleted, users?


brylie:
We have deleted several users from Kimai, and they show up when we click "Show deleted users" as expected. However, we would like to fully delete these users, or reactivate them, so that we can use their usernames. How can we fully delete users, or reactivate them?

ServiusHack:
Kimai currently does not offer this in the user interface.

Reactivating users could be easily added.

Fully deleting users is more problematic since timesheet entries still reference that user. So either we have timesheet entries without an associated user or these need to be deleted as well.

kevinpapst:
In the master branch isn now the new feature to reactivate "trashed" users.
Upon finally deleting the user we empty more db tables, but keep expenses and timesheet entries.

PR #597, opened 2016-03-04
Database refactoring (#566, #501, #469)

### LINKED ISSUE #501: Incomplete delete - users

### LINKED ISSUE #566: Remove unused variables

kevinpapst:
- removed Kimai_Database_Abstract, as we do not have multiple DB layers (anymore)
- removed server_type and server_connection as we do not support different database types
- adjusted write_config_file to not write server_type and server_conn anymore (shrinks autoconf.php)
- code style/reformatting/brackets in our Mysql class
- added auto-connect feature in Mysql class constructor
- rewritten user_get_config and customer_get_config internally to return data instead of merging it automatically into kga (this is now done in checkUserInternal)
- delete all possible data on user_delete (leave timesheet and expense entries)
- Fixed "Undefined index: showQuickNote" for new users and customers

Fixes #566
Fixes #501
Partially linked to #469 

Sorry, I know this is a huge PR, but I didn't know how to split it into smaller parts. Its a mixture of cleanup, bug fixing and refactoring ... you know: leave the code better than you found it.

It may be hard to review, I can just say: I tested it thoroughly and could not find any problems.


libraries/Kimai/Database/Mysql.php
```diff
@@ -1761,11 +2048,11 @@ public function configuration_edit($data) {
...
      */
-    public function get_current_recordings($userID) {
-
-        $p = $this->kga['server_prefix'];
+    public function get_current_recordings($userID)
+    {
+        $table = $this->getTimeSheetTable();
         $userID = MySQL::SQLValue($userID, MySQL::SQLVALUE_NUMBER);
-        $result = $this->conn->Query("SELECT timeEntryID FROM ${p}timeSheet WHERE userID = $userID AND start > 0 AND end = 0");
+        $result = $this->conn->Query("SELECT timeEntryID FROM $table WHERE userID = $userID AND start > 0 AND end = 0");
```
simonschaufi, line 2055: I don't like php variables in strings. I think even PSR doesn't allow them.

simonschaufi:
apart from some code style issue(s) it looks good.
code visually checked: :+1: 
code testet: no

libraries/Kimai/Database/Mysql.php
kevinpapst, line 2055: True, its almost like before ;-) yes we have to clear that in the future!

Issue #557, opened 2016-02-08
Not able to add Timesheet entry without setting rate for user


cljk:
As already pointed out by others the rates are a bit buggy - but I digged into a concrete problem for the current HEAD source tree.

If a user has no rate set, he is *not* able to add a timesheet entry. As I´ll try to show this has more than one reason.

`ki_timesheets/processor.php` is called with the field `rate` unset.... so far so clear. To initialize the form first `$database->get_best_fitting_rate` is called - there is the first error?!
In `Mysql.php` the function `get_best_fitting_rate` tries to estimate the best rate... if there are globals set (thats what I understand the function to do). But on my system it doesn´t. 

The query is produced correctly
```
      $query = "SELECT rate FROM " . $this->kga['server_prefix'] . "rates WHERE
    (userID = $userID OR userID IS NULL)  AND
    (projectID = $projectID OR projectID IS NULL)  AND
    (activityID = $activityID OR activityID IS NULL)
    ORDER BY userID DESC, activityID DESC, projectID DESC
    LIMIT 1;";

        $result = $this->conn->Query($query);
```
results to 
```
SELECT rate FROM kimai_rates WHERE (userID = 535744355 OR userID IS NULL) AND (projectID = 2 OR 
 projectID IS NULL) AND (activityID = 4 OR activityID IS NULL) ORDER BY userID DESC, activityID DESC, projectID 
 DESC LIMIT 1
```

I´ve filled a rate for the project (ID 2) previously and if I query it manually I get a single result as expected: "rate = 1.00"
BUT - (BUG) in `Mysql.php`on line 2889 the check for `RowCount`
```
if ($this->conn->RowCount() == 0)
            return false;

        $data = $this->conn->rowArray(0,MYSQLI_ASSOC);
        return $data['rate'];
```
returns "-1" instead of "1" - so the flow continues and an assoc array is tried to be fetched - but instead of an assoc array "false" is assigned to `$data`.
BTW $result->num_rows IS "1". (result is of type mysqli_result).

So obviously the query is executed correctly and the mysql-server returns a row - but the mapper class doesn´t work correctly here.

I tracked it down to `mysql.class.php`
```
$result = @mysqli_affected_rows($this->mysql_link);
			if (! $result) {
				$this->SetError();
				return false;
			} else {
				return $result;
			}
```
I don´t understand why but `@mysqli_affected_rows($this->mysql_link);` does indeed return "-1" and is the source for the confusion.

[...]

PR #599, opened 2016-03-05
Hotfix for #557 - cannot create time entry without rate

kevinpapst:
This is a bugfix for #557 

It incorporates changes inspired by @cljk https://github.com/cljk/kimai/commit/43c3c936c7175c19747ed88b852f305df66f5ed8
- return 0.00 as best fitting rate if none could be found in the database
- display an error if a new timesheet entry could not be created (currently the floater closes silently) and keep the floater visible
- set the rate to 0.00 in Mysql class if not present, as safety caution


simonschaufi:
TODO: search for rate set in preferences for example or somewhere else.

PR #602, opened 2016-03-05
Add function to restore a "trashed" user

### LINKED ISSUE #469: How to fully delete, or reactivate deleted, users?

kevinpapst:
FIXES #469

This PR adds the functionality to restore a "trashed" user.
Currently we can only fully delete these users, now we can restore them as well (if they were deleted by accident).

Also done:
- Added a column to display the users global user in the users admin panel (currently we can only see the group memberships)
- Code style improvements "on the fly" ;-)
- Created a functions.php include for admin extensions, so the code is not 1:1 duplicated in init.php and processor.php
- Improved Kimai_View and using this class instead of Zend_View
- Improved some translations


extensions/ki_adminpanel/processor.php
```diff
@@ -307,14 +287,25 @@ function ($group) {
 
         if (count($errors) == 0) {
             switch ($axValue) {
-                case 1 :
-                    // If the confirmation is returned the user gets the trash-flag.
+                // If the confirmation is returned the user is moved to trash
```
simonschaufi: wrong comment?

language/de.php
```diff
@@ -337,8 +338,8 @@
         'customerWithSameName' => 'Benutzernamen dürfen nicht gleich lauten wie Kundennamen.',
         'sameGlobalRoleName' => 'Eine globale Rolle mit diesem Namen existiert bereits.',
         'sameMembershipRoleName' => 'Eine gruppenbezogene Rolle mit diesem Namen existiert bereits.',
-        'noActivitySelected' => 'Keine Aktivität ausgwählt.',
-        'noProjectSelected' => 'Kein Projekt ausgwählt.',
+        'noActivitySelected' => 'Keine Aktivität ausgewählt.',
```
simonschaufi, line 341: hey, that was schwäbisch :D

extensions/ki_adminpanel/processor.php
kevinpapst: detective simon spotted a copy&paste mistake ;-) i'll push an update after you are done!

language/de.php
kevinpapst, line 341: :+1:

kevinpapst:
Updated

PR #605, opened 2016-03-06
Permission bugfixes

kevinpapst:
There were some "hidden" problems in Kimai, when working with none admin users:
- a user without the "role membership permission" > "regular user" > "view" could see all users in the system (thanks @Nef10) 
- customers got to see no users after refreshing the sublist "users", which was initially filled correctly
- when populating the select box for "allUser" a wrong parameter was passed (due to PHPs forgiving nature, it did not lead to a bug here)

### DIFF
--- a/core/processor.php
+++ b/core/processor.php
@@ -161,15 +161,15 @@ switch ($axAction) {
     break;
 
     /**
-     * Return a list of users. Customers are not shown any users. The
-     * type of the current user decides which users are shown to him.
-     * See get_user_watchable_users.
+     * Return a list of users.
+     * The type of the current user decides which users are shown to him.
+     * @see kimai.php as well
      */
     case 'reload_users':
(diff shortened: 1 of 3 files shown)


kevinpapst:
@simonschaufi If you have the chance, please try to reproduce the explained problems (see linked issue) in your local environment (if you have enough test data/user). 
I could reproduce the problem and my fixes seem to work, but as you know, the permissions are a "little confusing" and I am not sure if the submitted patch only does what it is intended to do.

simonschaufi:
works now :+1: 
for me it is still confusing whats the diff between global roles and membership roles. the documentation is really bad in this area. needs to be improved!

kevinpapst:
For me as well, but while I worked on that bug yesterday, I got a little bit of a better understanding.
Though, the implementation in the codebase is really nasty, that needs to be improved together with a documentation.

PR #608, opened 2016-03-07
Comment types

kevinpapst:
- removing duplicated code
- new view helper to render the comment type select-box
- remove $commentTypes from global namespace

- code cleanup

### DIFF
--- a/templates/helpers/CommentTypeSelect.php
+++ b/templates/helpers/CommentTypeSelect.php
@@ -24,14 +24,12 @@
  */
 class Zend_View_Helper_CommentTypeSelect extends Zend_View_Helper_FormSelect
 {
+    /**
+     * @param null $value
+     * @return string
+     */
     public function commentTypeSelect($value = null)
     {
-        $commentTypes = array(
-            0 => $this->view->translate('ctype0'),
-            1 => $this->view->translate('ctype1'),
-            2 => $this->view->translate('ctype2')
-        );
-
         return $this->formSelect(
             'commentType',
             $value,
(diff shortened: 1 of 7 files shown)


templates/helpers/CommentTypeSelect.php
```diff
@@ -0,0 +1,56 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Kimai; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * A select box to choose the comment type.
+ *
+ * @author Kevin Papst
```
simonschaufi, line 23: i would remove this

templates/helpers/CommentTypeSelect.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class Zend_View_Helper_CommentTypeSelect extends Zend_View_Helper_FormSelect
+{
+    /**
+     * @param null $value
```
simonschaufi: type is string or int, not null

templates/helpers/CommentTypeSelect.php
kevinpapst, line 23: we have to discuss this. its all over the code and i don't mind this for new files (just find it a bit overwhelming for each method, like in mysql class).
I am going to start a section in the documentation about coding standards ...

simonschaufi, line 23: what do you want to achieve with this? git can show me the author of every line. we don't need that extra information.

kevinpapst, line 23: credits to whom they belong. besides that: why not?

simonschaufi, line 23: we are a team, not single devs. this is **our** code ;)

PR #619, opened 2016-03-17
Autofocus for customer/project/activity floater

### LINKED ISSUE #522: Focus cursor on input field #name in popup (creating client/project/task)
In the WebClient:

### Now
- in the timesheet view, 
- clicking the plus symbol (at the very bottom) to add e.g a new client 
- a popup opens up to create one. 
- One moves the mouse cursor from the bottom of screen to the input field (middle) of the popup to give it a name.

### Suggestion
make the focusing part automatic. 
- clicking +
- popups opens
- cursor (for typing) is in the name input field of the popup

Obviously this would be a (little) time saver and i think would improve the user experience,
saving the mouse moving, to allow faster creation.

Additional benefit: if the focus would be on the name input field, the other fields (except those behind tabs) could be reached by the tabulator key, allowing fast creation of new entries. (type name>tab key>type etc, hit enter, be done)

### some quick observations
it looks like every of those name input fields can be identified by #floater fieldset#general input#name

### related thoughts on expanding the idea
In an ideal scenario, the tabs (in the popup, or maybe in whole gui) could be reached by keyboard keys as well. Maybe the whole gui should be capable of being shortcut driven (in a helpful thoughtout way), i'm not sure what would be a good implementation of such features and its maybe a topic for another issue.

## posibly helpful article (use ARIA)
- http://www.smashingmagazine.com/2015/12/making-accessibility-simpler/

kevinpapst:
FIXES #522

- auto-focus name input when "create new" floater opens

### DIFF
--- a/templates/scripts/core/main.php
+++ b/templates/scripts/core/main.php
@@ -359,7 +359,7 @@
 
 <div id="customers_foot">    
 <?php if ($this->show_customer_add_button): ?>
-        <a href="#" class="addLink" onclick="floaterShow('floaters.php','add_edit_customer',0,0,450); $(this).blur(); return false;"></a>
+        <a href="#" class="addLink" onclick="floaterShow('floaters.php','add_edit_customer',0,0,450,function() {$('#floater input#name').focus();}); $(this).blur(); return false;"></a>
 <?php endif; ?>
 <a href="#" class="selectAllLink" onclick="lists_filter_select_all('customers'); $(this).blur(); return false;"></a>
 <a href="#" class="deselectAllLink" onclick="lists_filter_deselect_all('customers'); $(this).blur(); return false;"></a>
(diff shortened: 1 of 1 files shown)


templates/scripts/core/main.php
```diff
@@ -359,7 +359,7 @@ function kimai_onload() {
 
 <div id="customers_foot">    
 <?php if ($this->show_customer_add_button): ?>
-        <a href="#" class="addLink" onclick="floaterShow('floaters.php','add_edit_customer',0,0,450); $(this).blur(); return false;"></a>
+        <a href="#" class="addLink" onclick="floaterShow('floaters.php','add_edit_customer',0,0,450,function() {$('#floater input#name').focus();}); $(this).blur(); return false;"></a>
```
simonschaufi: the jQuery selectors should be optimized: `$('#floater').find('#name')`.
there must always one one element of an id, so the tag name is not necessary. If we are sure there is only one id with value "name" when we can even shorten it more.

kevinpapst: True with the one id selector, but I am absolutely unsure if we have only one id="name" on the page - especially after all extensions are loaded - so I added a safer selector.
From what I read, there is no difference in one selector or splitting it up in one selector and one find() - but to be honest I just read a couple of Stackoverflow threads. Do you have a link with more info about that optimization? My javascript is clearly a bit rusty... ;-)

simonschaufi: - https://learn.jquery.com/performance/optimize-selectors/#id-based-selectors
- http://stackoverflow.com/questions/3177763/what-is-the-fastest-method-for-selecting-descendant-elements-in-jquery
- http://vaughnroyko.com/the-real-scoop-on-jquery-find-performance/
- and lots [more](https://www.google.com/search?q=jquery+selector+find+faster)...

kevinpapst: Thanks, you are the best! 
Even though I can't believe that we will recognize any difference ;-) your optimized version seems to be the prefered one.

kevinpapst:
Updated

templates/scripts/core/main.php
simonschaufi: thank you for your compliment :heart_eyes:
raw javascript would still be a lot faster but we will leave it for now.

kevinpapst:
Hm, we should add the same feature for the "amount" field of the "add expense" floater.

PR #620, opened 2016-03-17
Fix floater title

kevinpapst:
- switch title depending on whether a new or existing entry is opened

- while migrating to Zend_View, we forgot to check the $id on $this => regression bug

### DIFF
--- a/templates/scripts/floaters/add_edit_activity.php
+++ b/templates/scripts/floaters/add_edit_activity.php
@@ -1,6 +1,12 @@
 <div id="floater_innerwrap">
     <div id="floater_handle">
-        <span id="floater_title"><?php if (isset($id)) echo $this->kga['lang']['edit'], ': ', $this->kga['lang']['activity']; else echo $this->kga['lang']['new_activity']; ?></span>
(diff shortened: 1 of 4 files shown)


simonschaufi:
I was already working on exactly the same. Now you were faster ;)

templates/scripts/floaters/add_edit_activity.php
```diff
@@ -1,6 +1,12 @@
 <div id="floater_innerwrap">
     <div id="floater_handle">
-        <span id="floater_title"><?php if (isset($id)) echo $this->kga['lang']['edit'], ': ', $this->kga['lang']['activity']; else echo $this->kga['lang']['new_activity']; ?></span>
+        <span id="floater_title"><?php
+            if (isset($this->id) && $this->id !== 0) {
+                echo $this->kga['lang']['edit'], ': ', $this->kga['lang']['activity'];
```
simonschaufi: string concatenation with dot!

kevinpapst:
Sorry for stealing the fix ;-) 
I just stumbled upon that and could not look away.
Updated the dots.

PR #621, opened 2016-03-17
Budget improvements

### LINKED ISSUE #78: To many Tasks in "Budget"
Hey there, the Budget-Extension shows each possible Task, but I have an Project with just 4 tasks, Kimai shows me 10 ^^ That is too much.

Can you add an filter?

kevinpapst:
FIXES #78

Please test it locally with some productive data... 

1. Budget extension was quite slow and heavily overloaded with JS and Canvas
2. After this PR we only render charts that display information, no empty and useless plots any longer

- do not show projects without any data
- do not show tasks without any data
- improved box title: include customer for overview
- translated labels
- using <table> for data display (UI improvements including some css updates)
- code cleanup & reformatting
- use Kimai View instead of Zend View


simonschaufi:
- used my own live data and I still see all the projects with budget of zero. Shouldn't they be invisible? I would expect to see only graphs of projects/activities that have at least a budget set. I hope you can explain me what approved and effort means ;)
- budget_plot_data function is way to complex. do you want to extract some foreach loops in other methods?
- I also optimized this PR quite a lot. I thought from now on we use devel branch or are we still waiting for something? Then I could merge my own things easily without breaking master. I checked if it is possible for me to add more commits to your PR but thats not possible.

extensions/ki_budget/templates/scripts/charts.php
```diff
@@ -1,65 +1,128 @@
...
+{
+    if (array_search($project['projectID'], $this->projects_selected) === false) {
+        continue;
+    }
 
-<?php endforeach; ?>
+    $temp = $project['projectID'];
+
+    // do not render projects that have only empty values
+    if ($this->arr_plotdata[$temp]['total'] == 0 &&
```
simonschaufi: this should be done in processor.php where data is fetched, not in the view

extensions/ki_budget/templates/scripts/charts.php
```diff
@@ -1,65 +1,128 @@
...
+        $this->arr_plotdata[$temp]['budget'] == 0 &&
+        (!isset($this->arr_plotdata[$temp][0]['expenses']) || $this->arr_plotdata[$temp][0]['expenses'] == 0)) {
+        continue;
+    }
```
simonschaufi: this should be done in processor.php where the data is fetched, not in the view

extensions/ki_budget/templates/scripts/charts.php
```diff
@@ -1,65 +1,128 @@
...
+        if (array_search($id, $this->activities_selected) === false) {
+            continue;
+        }
+
+        if ($activity['total'] == 0 &&
+            $activity['budget'] == 0 && $activity['budget_total'] == 0 &&
+            $activity['approved'] == 0 && $activity['approved_total'] == 0
+        ) {
+                continue;
+        }
```
simonschaufi: likewise here

extensions/ki_budget/templates/scripts/charts.php
```diff
@@ -1,65 +1,128 @@
...
+            </table>
+        </div>
+        <?php
+    }
+    ?>
+    <br>
+    <?php
+}
+?>
 <div class="budget_project_end"/>
```
simonschaufi: there is a html error. divs cannot be self closed.

extensions/ki_budget/templates/scripts/charts.php
```diff
@@ -1,65 +1,128 @@
...
-//		} catch(e) {
-//			alert(e);
-//		}
-     });
-    </script>
+        chartColors = <?php echo $this->chartColors ?>;
+        budget_extension_plot(<?php echo $this->javascript_arr_plotdata ?>);
+        recalculateWindow();
+    });
+</script>
```
simonschaufi: please move this to the bottom of the file.

extensions/ki_budget/processor.php
```diff
@@ -25,152 +25,165 @@
...
+        //	$view->projects = $projects;
+        //	$view->activities = $activities;
+        //}
+        //else {
+        //	$view->projects = 0;
+        //}
+        //$chartColors = array("#efefef", "#4bb2c5", "#EAA228", "#c5b47f", "#579575", "#839557", "#958c12", "#953579", "#4b5de4", "#d8b83f", "#ff5800", "#0085cc");
+        //$view->chartColors = json_encode($chartColors);
+        //
+        //        echo json_encode($arr_plotdata);
```
simonschaufi: can this code block (comment) be removed?

simonschaufi:
- js function recalculateWindow should be moved from index.php into budget_func.js as there is no php code rendered into the function.
- in index.php there should just be this tag: `<div id="budgetArea"></div>` without further nested content. Instead the loader should be visible until the charts are completely rendered.
- ajax request should show the loader and success function should hide it again.

kevinpapst:
Simon, I wish more developer would do code reviews like you, fantastic feedback!

1) I thought we would use develop after the next release ... we can start right after this PR is done. Or as an alternative, I can close the PR and re-open it against the develop branch.

2) This is the first time I was working on the budget extension , so I am not sure about all the logic in the background. I guess it must be refactored like all the rest ;-) No clue what approved/effort means or how it works.

3) The main idea behind this PR was to remove all plots/charts that were completely empty. With this PR you should now ONLY see plots that have data. Not only projects that have a budget set, but any project/activity with data (either 1. timesheet entries or 2. expenses or 3. projects with budgets > 0).

Conclusion: this PR does only remove completely empty plots (which were about 80% in my production data). The amount of empty data is/was highly depending on your tasks configuration. If you use exclusive tasks for projects, less plots should be empty. But if you have a lot of unassigned tasks, these tasks were rendered for each and every project and almost every one had no data => empty plot.

I pushed another try ;-) could you have a look again?!

This one brings in a new CSS file, so extensions do not have to hard link images from the standard skin!

PR #623, opened 2016-03-19
Added unit tests

kevinpapst:
- Added some more unit tests

### DIFF
--- a/tests/library/Kimai/ViewTest.php
+++ b/tests/library/Kimai/ViewTest.php
@@ -22,8 +22,6 @@ namespace KimaiTest;
 use Kimai_View;
 
 /**
- * Class UserTest
- *
  * @coversDefaultClass Kimai_View
  */
 class ViewTest extends TestCase
--- a/tests/library/Kimai/ViewTest.php
+++ b/tests/library/Kimai/ViewTest.php
@@ -19,20 +19,36 @@
 
 namespace KimaiTest;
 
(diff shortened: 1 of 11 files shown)


tests/library/Kimai/ViewTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->assertArrayHasKey('Zend_View_Helper_', $helperPaths);
+        $this->assertContains(APPLICATION_PATH . '/templates/helpers/', $helperPaths['Zend_View_Helper_']);
+
+        $scriptsPaths = $view->getScriptPaths();
+        $this->assertContains(APPLICATION_PATH . '/templates/scripts/', $scriptsPaths);
+
+        $vars = $view->getVars();
+        $this->assertArrayHasKey('kga', $vars);
+
+        $this->assertEquals($myKga, $vars['kga']);
```
simonschaufi, line 50: do you think we really need to test this? Imho we don't need to test Zend internal methods (`getVars`). When we assign variables to the template, we are sure they are there.

kevinpapst, line 50: True. But the test just makes sure:
- that "kga" is imported from the global namespace and
- that nobody accidentally removes "kga" from the view, because almost every template uses it.

simonschaufi, line 50: ok.

PR #628, opened 2016-03-20
Followup to #621

simonschaufi:
- Truncate strings in charts
- move label box a bit more to the left to avoid overlapping

### DIFF
--- a/extensions/ki_budget/js/budget_func.js
+++ b/extensions/ki_budget/js/budget_func.js
@@ -21,49 +21,49 @@
 // ==========================
 
 function budget_extension_onload() {
-    budget_extension_resize();
-    $("#loader").hide();
-    lists_visible(true);
-    try {
-    budget_extension_reload();
-    } catch(e) {
-    	alert(e);
-    }
+	budget_extension_resize();
+	$('#loader').hide();
+	lists_visible(true);
+	try {
+		budget_extension_reload();
+	} catch(e) {
+		alert(e);
+	}
 }
 
 function recalculate_budget_window() {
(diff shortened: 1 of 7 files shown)


kevinpapst:
30px right for the label box moves it almost into the chart (chrome on mac). doesn't look good here, because there is a lot of space left between the chart and the box border. I guess its highly depending on usernames and translation of the label ... any idea? JS positioning?

image

Truncate is a good idea indeed.

simonschaufi:
try it with long usernames like simonschaufi. the original position then overlapped

extensions/ki_budget/js/budget_func.js
```diff
@@ -72,133 +72,136 @@ function recalculate_budget_window() {
  */
 function budget_extension_set_heightTop() {
 	budget_extension_get_dimensions();
-    if (!extensionShrinkMode) {
-        $("#budgetArea").css("height", budget_h);
-    } else {
-        $("#budgetArea").css("height", "20px");
-    }
-    
-    export_extension_set_TableWidths();
```
simonschaufi, line 81: I didnt understand why this was there. maybe copy paste error, maybe mistake...

kevinpapst, line 81: Probably, or likely Copy&Paste. But I have no clue either. I work the same way all the time: change and see what happens. Probably we should set the label box to a fixed with and truncate the username?

kevinpapst:
Hey, we can merge that, I was just wondering if you are looking into another option? Probably javascript positioning, like the rest of the UI

simonschaufi:
let's make another PR for that...

PR #637, opened 2016-03-25
Login forms

kevinpapst:
This Pull requests fixes the completely broken warning box feature in the login screen.
Simply turn off javascript at the demo site and you will see what I mean.
- translated reset password button
- formatted code
- fixed warning (javascript, cookies) box
- using translate viewhelper
- do not check on startup in forgotPassword screen

### DIFF
--- a/processor.php
+++ b/processor.php
@@ -27,8 +27,8 @@ if (!isset($_REQUEST['a'])) {
 }
 
 if (!isset($_POST['name']) || is_array($_POST['name'])) {
-    $name = ""; 
-} else { 
+    $name = "";
+} else {
     $name = $_POST['name'];
 }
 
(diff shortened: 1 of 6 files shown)


templates/scripts/login/panel.php
```diff
@@ -70,53 +69,50 @@
     <link rel="stylesheet" type="text/css" media="screen" href="<?php echo $this->skin('login.css'); ?>" />
 </head>
 <body>
-  <div id="content">
-
+<div id="content">
     <div id="box">
+        <div id="login" style="display:block">
```
simonschaufi, line 74: please no inline css

templates/scripts/login/panel.php
```diff
@@ -70,53 +69,50 @@
...
+        <div id="forgotPasswordConfirmation">
+            <p></p>
+            <a class="returnToLogin" href=""><?php echo $this->translate('passwordReset:returnToLogin') ?></a>
+        </div>
             
-            <div id="warning">
-                <p id="JSwarning"><strong style="color:red"><?php $this->kga['lang']['JSwarning']?></strong></p>
-                <p id="cookiewarning"><strong style="color:red"><?php $this->kga['lang']['cookiewarning']?></strong></p>
-            </div>
+        <div id="warning" style="display:block">
```
simonschaufi, line 110: no inline css

templates/scripts/login/panel.php
kevinpapst, line 74: needs to be done here, otherwise I have to rewrite the complete css

skins/standard/login.css
```diff
@@ -229,10 +229,8 @@ body {
 }
 
 #warning {
-	margin: 5px 0 0 38px;
-	width: 434px;
 	border: 2px solid red;
-	padding: 10px 5px 5px 5px;
+	padding: 10px 5px 5px 5px !important;
```
simonschaufi, line 233: is important really needed?

processor.php
```diff
@@ -55,82 +55,79 @@
...
-            $message);
+            $ssl = !empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off';
+            $url = ($ssl ? 'https://' : 'http://') . $_SERVER['SERVER_NAME'] . dirname($_SERVER['SCRIPT_NAME']) . '/forgotPassword.php?name=' . urlencode($name) . '&key=' . $passwordResetHash;
 
-          echo json_encode(array(
-            'message' => $kga['lang']['passwordReset']['mailConfirmation']
-          ));
+            $message = $kga['lang']['passwordReset']['mailMessage'];
+            $message = str_replace('%{URL}', $url, $message);
+            mail(
```
simonschaufi, line 80: if you want to do further refactoring, use [Zend Mail](http://framework.zend.com/manual/1.12/de/zend.mail.html) here

templates/scripts/login/panel.php
kevinpapst, line 110: see above, also there is already inline css

skins/standard/login.css
kevinpapst, line 233: Otherwise i wouldn't have added it here ;-) 

I guess someone updated the login.css in the past and completely forgot to test the warning box.  It was totally broken... and I have no clue why the other classes were added that influence the warning box so heavily. 

This rule only applies to the two auth screens, so I don't care about the important here, which I would otherwise (in a global rule) not use.

But if you find another/better way, go ahead, merge and update it - as said in the past: I have little frontend knowledge.

processor.php
kevinpapst, line 80: Later, that PR is only about fixing the completely broken "warning box".

simonschaufi:
i tried to "reproduce" the problem but don't really see whats the issue here except that the forgot password does not work without javascript.

kevinpapst:
What I wrote ;-)
First problem:
- Deactivate Javascript and check if you see a "please activate javascript" error message. Hint: you won't see it. 
- Now give the div#warning the style "display:block" and enjoy the beautiful error messsage:

image

Second "problem":
- Missing Translation on the button

simonschaufi:
I merged it and will do a follow up on this as i cannot fork your fork of kimai

kevinpapst:
Very good idea! I thought that we are not done with the login forms yet, thats why I wanted to have it in develop and not master ... it is not urgent, as nobody complained ;-)

PR #640, opened 2016-03-25
Fix export table alignment

### LINKED ISSUE #105: Aligning field names with table
I thought this was just a problem with my translation file, but it is actually a general issue:
when in the tab 'Export', the field names are not aligned with the table.
I don't know what would be the best solution here. 
Maybe working with abbreviations?
Or having the word 'cut' when the length of a field is being exceeded (maybe having the full word appear on hover over)?
This is not really an urgent issue, but it should clarify this already dense tab.

21nqwdc

### LINKED ISSUE #622: UI Changed With Change of Time Format.
#### Expected behaviour
I edited extenstions/kiexport/init.php file to change date format from DD.MM to DD.MM.YYYY. However now I see table header is not showing properly with table column. You can see it here.. http://prnt.sc/ah4o4s. I just want to show the columns exactly underneath of every column header. From which file I can edit the HTML? Any help would be highly appreciated. 

#### Actual behaviour
Table headers are not showing over table columns.

I think the table is creating dynamically so changing the HTML would require changing in PHP/JavaScript codes too.

#### Kimai version
Latest version

kevinpapst:
FIXES #105 
FIXES #622  

Reason for this pull request:
- Wrong alignment of export table header

- align field names with table
- added view helper to truncate strings and added an ellipsis
- using ellipsis view helper to shorten translations of long table headers causing the wrong alignment
- using error view helper & removed unused translation string "noEntries"

ATTENTION: 
The main problem is fixed now, but I guess we won't be able to fix this problem ever on smaller screens. 
What is still there: if the header translations will be longer than the content columns, the header table will expand and then not properly aligned. This should be prevented by the ellipsis view helper used in every table header. But if the screen size shrinks too much, the table will be adjusted by the browser down to a point when that concept breaks and the alignment falls apart. 

But: the same is true for the timesheet, there are just less columns and therefor the problem is much smaller. I believe we should stick to the same logic in both tables and just accept, that Kimai is not made for small screens.


extensions/ki_export/js/xp_func.js
```diff
@@ -183,23 +183,26 @@ function export_extension_set_TableWidths() {
...
+	$("#export_head > table > tbody > tr > td.customer").css("width", 	$("div#xp > div > table > tbody > tr > td.customer").width());
+	$("#export_head > table > tbody > tr > td.project").css("width", 	$("div#xp > div > table > tbody > tr > td.project").width());
+	$("#export_head > table > tbody > tr > td.activity").css("width", 	$("div#xp > div > table > tbody > tr > td.activity").width());
+	$("#export_head > table > tbody > tr > td.description").css("width",$("div#xp > div > table > tbody > tr > td.description").width());
+	$("#export_head > table > tbody > tr > td.comment").css("width", 	$("div#xp > div > table > tbody > tr > td.comment").width());
+	$("#export_head > table > tbody > tr > td.location").css("width", 	$("div#xp > div > table > tbody > tr > td.location").width());
+	$("#export_head > table > tbody > tr > td.trackingNumber").css("width", $("div#xp > div > table > tbody > tr > td.trackingNumber").width());
+	$("#export_head > table > tbody > tr > td.user").css("width", 		$("div#xp > div > table > tbody > tr > td.user").width());
+	$("#export_head > table > tbody > tr > td.cleared").css("width", 	$("div#xp > div > table > tbody > tr > td.cleared").width() + scr);
```
simonschaufi, line 205: the worst jquery selectors ever but this should be fixed globally in every file

kevinpapst, line 205: LOL, I just used what was there ;-) but you are completely right!

PR #643, opened 2016-03-26
Hide all related projects if client is made invisible

### LINKED ISSUE #430: Hide a clients projects if client is made invisible
Currently you have to hide the client, then hide all of their related projects manually, ideally if a client is hidden all of their projects become hidden.

It would probably make sense for this not to require the checkboxes to be set on the projects, so that you could make the client visible again and any of their project checked as visible would be displayed again.

kevinpapst:
FIXES #430

- automatically hide projects, if the linked customer is hidden
- removed never touched code and silly html comments
- some phpdoc and code style changes on the fly
- show projects as hidden in admin if customer is hidden
- hide projects and customers in floater select boxes (unless they are linked in existing entries)

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -2,7 +2,7 @@
 /**
  * This file is part of
  * Kimai - Open Source Time Tracking // http://www.kimai.org
- * (c) 2006-2009 Kimai-Development-Team
+ * (c) Kimai-Development-Team since 2006
  *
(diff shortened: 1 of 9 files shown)


kevinpapst:
DO NOT MERGE YET! There is a problem in the administration area, where activities do not display the projects if those were hidden. I have to work on that first.

PR #648, opened 2016-03-29
fixed none-working export feature "print html table"

kevinpapst:
- fix the submit button in the "print html table" floater in the export extension
- print the html table with utf-8 charset to be able to render all translations properly
- some minor code style changes

Reason for this pull request:
- feature was not working at all (probably this was only defect in some browser due to the "return false" statement)

### DIFF
--- a/extensions/ki_export/templates/scripts/floaters/print.php
+++ b/extensions/ki_export/templates/scripts/floaters/print.php
@@ -37,7 +37,7 @@
                 </ul>
                 <div id="formbuttons">
                     <input class='btn_norm' type='button' value='<?php echo $this->kga['lang']['cancel'] ?>' onclick='floaterClose();return false;'/>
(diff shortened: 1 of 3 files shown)


extensions/ki_export/templates/scripts/floaters/print.php
```diff
@@ -37,7 +37,7 @@
                 </ul>
                 <div id="formbuttons">
                     <input class='btn_norm' type='button' value='<?php echo $this->kga['lang']['cancel'] ?>' onclick='floaterClose();return false;'/>
-                    <input class='btn_ok' type='submit' value='<?php echo $this->kga['lang']['submit'] ?>' onclick="floaterClose();return false;"/>
+                    <input class='btn_ok' type='submit' value='<?php echo $this->kga['lang']['submit'] ?>' onclick="floaterClose();"/>
```
simonschaufi, line 40: that bug was introduced by mistake

kevinpapst, line 40: I guess so, who would introduce bugs purposely? ;-)

Issue #601, opened 2016-03-05
View cleanup


kevinpapst:
Cleanup the usage of Zend_View
- replace all occurrences of Zend_View with Kimai_View
- replace $view->kga = $kga by changing the init() method and injecting $kga there
- fix all path settings (relates to #483)

kevinpapst:
Ongoing process in multiple pull requests

kevinpapst:
Moved to https://github.com/kimai/roadmap/issues/2

PR #651, opened 2016-03-31
use $view->assign()

simonschaufi:
- instead of assigning the variables directly into the view, use the api method
- remove $user assignment in checkuser as user variable is never used

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -21,7 +21,7 @@
  * =============================
  * = Floating Window Generator =
  * =============================
- * 
+ *
  * Called via AJAX from the Kimai user interface. Depending on $axAction
  * some HTML will be returned, which will then be shown in a floater.
  */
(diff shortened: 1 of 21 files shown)


simonschaufi:
there is more to come...

kevinpapst:
Why into master? Such big refactorings should go into develop

simonschaufi:
because it doesn't break anything ;) that will be the last thing that goes into master. from now on i will commit into develop.

kevinpapst:
I have multiple branches at home, that rework on several aspects, including the View. 
I already see hundreads of merge conflicts... I don't think its a necessary change for master and would highly prefer if you update the PR to point to develop.

simonschaufi:
mhh, how can I know that you are working on the view as well? even if i point it to develop, that won't solve the issue of merge conflicts

kevinpapst:
https://github.com/kimai/kimai/issues/601#issuecomment-196999116

But more importantly: does that fix anything beside your naging IDE comments about using "magic setters" ? ;-)
I would like to keep such huge refactorings out of the way for the next release.

kevinpapst:
As discussed => merge into develop after fixing the missing empty lines in license header

PR #655, opened 2016-04-01
use $view->assign()

simonschaufi:
- instead of assigning the variables directly into the view, use the api method where possible (some assignments are as array which I didn't touch yet)

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -21,7 +21,7 @@
  * =============================
  * = Floating Window Generator =
  * =============================
- *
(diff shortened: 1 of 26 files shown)


kevinpapst:
No Unit tests for that? ;-) just kidding

I assume you tested everything and checked your log file. Visual check seems to be okay, but my eyes are not a good indicator for the validity of some many changes

simonschaufi:
yes of course, i checked every single line ;)

PR #656, opened 2016-04-01
FIX: undefined index defaultLocation

simonschaufi:
Reason for this pull request:
- undefined index defaultLocation


libraries/Kimai/Database/Mysql.php
```diff
@@ -3617,7 +3618,7 @@ public function startRecorder($projectID, $activityID, $user, $startTime)
             $values['rate'] = $rate;
         }
 
-        if (!$this->kga['conf']['defaultLocation'] == '') {
+        if (isset($this->kga['conf']['defaultLocation']) && !$this->kga['conf']['defaultLocation'] == '') {
```
kevinpapst, line 3621: This change is not necessary (I guess).
If you set defaultLocation in get_global_config(), then it will always be available?!?

PR #660, opened 2016-04-02
Remove date.js

simonschaufi:
this file is not used.

Original source: https://github.com/vitch/jquery-methods/blob/master/date.js


kevinpapst:
Weird that we have so many unused files...

simonschaufi:
i wonder if that has been in use already... it took me quite a while to :eyes: the original source of this file

PR #661, opened 2016-04-02
Update cookie.js (remove unmaintained version)

simonschaufi:
Source: https://github.com/js-cookie/js-cookie


templates/scripts/core/main.php
```diff
@@ -23,7 +23,7 @@
     <script src="../libraries/jQuery/jquery.hoverIntent.minified.js" type="text/javascript" charset="utf-8"></script>
     <script src="../libraries/jQuery/jquery.form.js" type="text/javascript" charset="utf-8"></script>
     <script src="../libraries/jQuery/jquery.newsticker.pack.js" type="text/javascript" charset="utf-8"></script>
-    <script src="../libraries/jQuery/jquery.cookie.js" type="text/javascript" charset="utf-8"></script>
+    <script src="../libraries/jQuery/js.cookie-2.1.0.min.js" type="text/javascript" charset="utf-8"></script>
```
kevinpapst, line 26: Should we move it to a "composer like" path /libraries/js-cookie/js.cookie-2.1.0.min.js ?

simonschaufi, line 26: i thought about it as well but then said no because it has no advantage. but we should write it down somewhere where we have the original source of the files from.

kevinpapst, line 26: What about here: https://github.com/kimai/kimai/blob/master/templates/scripts/floaters/credits.php#L60

PR #666, opened 2016-04-05
Green update from 0.9.2

simonschaufi:
make smooth upgrade from 0.9.3 (fresh install)

FIX:
- path to installPermissions
- all issues with keys

Related: #517 but does not solve the issues mentioned in this ticket!

### DIFF
--- a/updater/updater.php
+++ b/updater/updater.php
@@ -587,6 +587,8 @@ if ((int)$revisionDB < 1305) {
     }
 }
 
+// release of kimai 0.9.2 with r1306
+
 if ((int)$revisionDB < 1326) {
     Kimai_Logger::logfile("-- update to r1326");
     exec_query("INSERT INTO ${p}var (`var`,`value`) VALUES('editLimit','-')");
@@ -936,7 +938,7 @@ if ((int)$revisionDB < 1373) {
 
 if ((int)$revisionDB < 1374) {
     Kimai_Logger::logfile("-- update to r1374");
-    require("installer/installPermissions.php");
+    require("../installer/installPermissions.php");
 
     // add membershipRoleID column, initialized with user role
     exec_query("ALTER TABLE `${p}groups_users` ADD `membershipRoleID` int(10) DEFAULT $membershipUserRoleID;");
@@ -1079,6 +1081,8 @@ if ((int)$revisionDB < 1384) {
     exec_query("ALTER TABLE ${p}customers ADD COLUMN `passwordResetHash` char(32) NULL DEFAULT NULL AFTER `password`");
 }
 
+// release of kimai 0.9.3
+
 if ((int)$revisionDB < 1385) {
     Kimai_Logger::logfile("-- update to r1385");
     exec_query("ALTER TABLE ${p}customers CHANGE `comment` `comment` TEXT NULL;");
@@ -1111,6 +1115,7 @@ if ((int)$revisionDB < 1387) {
     exec_query("INSERT INTO `${p}configuration` (`option`,`value`) VALUES('date_format_3','d.m.Y')");
 }
 
+// release of kimai 1.0
 
 // ================================================================================
 // FINALIZATION: update DB version number
--- a/updater/updater.php
+++ b/updater/updater.php
@@ -106,9 +106,7 @@ if ((int)$revisionDB < 1219 && !isset($_REQUEST['timezone']))
     );
 }
 
-// ================================================================================
 require_once 'update_header.php';
-// ================================================================================
 
 $errors = 0;
 $executed_queries = 0;
@@ -133,7 +131,7 @@ if ((int)$revisionDB < $kga['revision']) {
     $query = ("SHOW TABLES;");
 
     $result_backup = $database->queryAll($query);
-    Kimai_Logger::logfile($query, $result_backup);
+    Kimai_Logger::logfile($query);
     $prefix_length = strlen($p);
 
     echo "</table>";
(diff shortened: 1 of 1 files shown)


updater/updater.php
```diff
@@ -925,24 +924,21 @@
 DROP `effort`,
 DROP `approved`
 ;", false);
-
-    exec_query("ALTER TABLE `${p}projects_activities`
-ADD `budget`     decimal(10,2) NOT NULL DEFAULT '0.00',
-ADD `effort`     decimal(10,2) DEFAULT NULL,
-ADD `approved`   decimal(10,2) DEFAULT NULL
-;", false);
```
simonschaufi, line 933: this was removed on purpose!

updater/updater.php
```diff
@@ -644,7 +640,7 @@
 
     $result = $database->queryAll("SELECT usr_ID,usr_grp FROM ${p}usr");
     foreach ($result as $row) {
-        exec_query("INSERT INTO ${p}grp_usr (`grp_ID`,`usr_ID`) VALUES($row[usr_grp],$row[usr_ID]);");
+        exec_query('INSERT INTO ' . $p . 'grp_usr (`grp_ID`, `usr_ID`) VALUES(' . $row['usr_grp'] . ',' . $row['usr_ID'] . ');');
```
kevinpapst, line 643: please follow the same pattern as all other statements

updater/updater.php
kevinpapst, line 933: Why? From the comment I read that the columsn will purposely be removed and then re-added.
Now they are only removed... please explain :)

updater/updater.php
```diff
@@ -925,24 +924,21 @@
...
 
 if ((int)$revisionDB < 1373) {
+    Kimai_Logger::logfile("-- update to r1373");
     exec_query("ALTER TABLE `${p}activities` DROP `assignable`;");
 }
 
 if ((int)$revisionDB < 1374) {
-    require("installer/installPermissions.php");
+    Kimai_Logger::logfile("-- update to r1374");
+    require("../installer/installPermissions.php");
```
kevinpapst, line 941: can we use a **DIR** constant here, to make 100% sure it will be loaded?

updater/updater.php
```diff
@@ -1113,6 +1131,4 @@
 
 Kimai_Logger::logfile("-- update finished --------------------------------");
 
-// ================================================================================
```
kevinpapst, line 1116: Where is all the ASCII art gone to? ;-)

updater/updater.php
```diff
@@ -1017,40 +1019,53 @@
...
+    exec_query("ALTER TABLE `${p}users` ADD UNIQUE  KEY(`apiKey`);", false);
 
     exec_query("UPDATE `${p}preferences` SET `option` = 'ui.project_comment_flag' WHERE `option` = 'ui.pct_comment_flag';");
 }
 
 if ((int)$revisionDB < 1382) {
     Kimai_Logger::logfile("-- update to r1382");
-    exec_query("ALTER TABLE `${p}membershipRoles` ADD `core-user-view` tinyint DEFAULT 0 AFTER `core-user-unassign`;", false);
+    #column already added in installer/installPermissions.php in r1374
+    #exec_query("ALTER TABLE `${p}membershipRoles` ADD `core-user-view` tinyint DEFAULT 0 AFTER `core-user-unassign`;", false);
```
kevinpapst, line 1068: Yes, but installPermissions is not executed in this step! And if it was added later on, people will miss this column.

updater/updater.php
```diff
@@ -1017,40 +1019,53 @@
 if ((int)$revisionDB < 1381) {
     Kimai_Logger::logfile("-- update to r1381");
     // make sure all keys are defined correctly
-    exec_query("ALTER TABLE `${p}activities`          ADD PRIMARY KEY(`activityID`);", false);
-    exec_query("ALTER TABLE `${p}configuration`       ADD PRIMARY KEY(`option`);", false);
-    exec_query("ALTER TABLE `${p}customers`           ADD PRIMARY KEY(`customerID`);", false);
-    exec_query("ALTER TABLE `${p}expenses`            ADD PRIMARY KEY(`expenseID`);", false);
+    # primary key since r733
```
kevinpapst, line 1022: This line and following: why did you remove them?

updater/updater.php
simonschaufi, line 933: They are removed from the **activities** table and added on the **projects_activities** table. Thats why this migration will throw an error on duplicate columns since they exist since [r1368](https://github.com/kimai/kimai/blob/master/updater/updater.php#L813-L821) already. I didn't find the revision where these columns should have been removed.

updater/updater.php
simonschaufi, line 1022: made a comment above: primary key since r733. the columns just get renamed and therefor still have the primary keys.

updater/updater.php
simonschaufi, line 1116: too much of it makes the code ugly :) For the footer it is really not needed.

updater/updater.php
simonschaufi, line 1068: installPermissions is run within one release (0.9.3). I made sure it is not run in 0.9.2. So this column gets added twice in 0.9.3 and if people are always on master, they already have it ;) So this can be safely removed.

updater/updater.php
simonschaufi, line 941: can you modify my pull request and add it? you now have the permission to do so ;)

updater/updater.php
simonschaufi, line 643: I actually want to get rid of variables in strings because of [this](https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#strings-and-concatination). I know this is not PSR-2 but it just makes the code more readable imho.

PR #667, opened 2016-04-09
set_include_path: Not delete system include path

HenryNe:
FIXES
Warning: parse_ini_file(config.ini) [function.parse-ini-file]: failed to open stream: No such file or directory in /srv/www/htdocs/kimai-0.9.3/includes/kspi.php on line 48

- set_include_path should not delete system default include paths.

Reason for this pull request:
- In my case my path ".:/usr/share/php:/usr/share/pear" was dropped by the old call of set_include_path.

### DIFF
--- a/includes/basics.php
+++ b/includes/basics.php
@@ -30,6 +30,7 @@ set_include_path(
     implode(
         PATH_SEPARATOR,
         array(
+            get_include_path(),
             realpath(APPLICATION_PATH . '/libraries/zendframework/zendframework1/library/'),
         )
     )


kevinpapst:
Hi @HenryNe -
could you please tell me, what went wrong in the first place and how that line could this fix your problem?

I know that your code, which was already merged (why?) causes major troubles on (at least) some systems. A lot of (especially older shared host) systems have still PEAR and ZF1 pre-installed and break composer autoloading if the system include path is at the first place the first in Kimai's include path. 

Using system include path is IMO not a good idea, as you can never tell if the correct file was loaded.
And I am pretty sure, that we had troubles with it before. Thats why we are shifting all of our efforts towards composer.

Would you share some insights on your environment and what went wrong in the first place?

And btw: Thanks for contributing :-) 
I just want to make sure that Kimai runs on as many hosts as possible!

simonschaufi:
I merged it because it sounded obvious and I couldn't really test it as I don't know how to reproduce it but if this fix fixes the problem for him, then it should be fine.

HenryNe:
My system is an old Debian 5.0.8 Lenny with PHP Version 5.2.6-1+lenny16, running lighttpd/1.4.19 with FastCGI as webserver.

The system include_path is ".:/usr/share/php:/usr/share/pear". I have detected with phpinfo().

I installed kimai version 0.9.3.1384, changed some unnamed function calls for array_filter, array_map and array_replace_recursive, and the version check from PHP 5.3 to PHP 5.2. I was able to run the Login as admin.

With a click at tapstop "Budget" or "Export" I got no "Projects" in the inner frame and this error message:
`2016-04-10 13:20:57: (mod_fastcgi.c.2610) FastCGI-stderr: PHP Warning:  parse_ini_file(config.ini) [<a href='function.parse-ini-file'>function.parse-ini-file</a>]: failed to open stream: No such file or directory in /srv/www/kimai/kimai-0.9.3/includes/kspi.php on line 40`

Via try and error I detected the `set_include_path` as failed myst.
Henry

HenryNe:
I played some more with that myst and found, that current directory is missed to fix my problem:

```
set_include_path(
    implode(
        PATH_SEPARATOR,
        array(
            '.',
            realpath(APPLICATION_PATH . '/libraries/'),
        )
    )
);
```

simonschaufi:
Let me ask a stupid question but why are you still using php 5.2?

HenryNe:
An upgrade would break other running web applications.

HenryNe:
The Path `.:/usr/share/php:/usr/share/pear` is compiled in in the binary php-cgi.
In /etc/php5/cgi/php.ini the include_path is not set.

kevinpapst:
Thats what I expected. We should change the previously added include path back to the minimal viable solution. I am sure, that the default include path can lead to problems.
Thanks for your feedback @HenryNe !

simonschaufi:
sooner or later we should get rid of the ini files alltogether.

PR #668, opened 2016-04-09
prevent possible "undefined index" in filters

HenryNe:
- Use empty() instead string compair and/or isset()

Reason for this pull request:
- Supress warnings about "undefined index" in filters

PS: Sorry, I don'k know, how to strip changes on set_include_path from that pull request. So, that is included here again.


includes/basics.php
```diff
@@ -30,6 +30,7 @@
     implode(
         PATH_SEPARATOR,
         array(
+            get_include_path(),
```
simonschaufi, line 33: this should definitely be removed from the PR.

HenryNe, line 33: I Have merged from kimai/devel into my branch HenryNe-general-bugfix.
In my view have no changes on file basics.php.
I don't know. Must I re-pull the pull request #668 ?

PR #669, opened 2016-04-09
Update credits.php

simonschaufi:
- Add more source links
- code style changes
- add external linking

### DIFF
--- a/templates/scripts/floaters/credits.php
+++ b/templates/scripts/floaters/credits.php
@@ -8,7 +8,7 @@
     <div class="floater_content" style="margin:10px">
         <h2>Kimai - Open Source Time Tracking</h2> 
         <p>
-            <?php echo 'v', $this->kga['version'],'.',$this->kga['revision'],' - &copy; ',$this->devtimespan;?> by the Kimai-Team:
+            <?php echo 'v' . $this->kga['version'] . '.' . $this->kga['revision'] . ' - &copy; ' . $this->devtimespan;?> by the Kimai-Team:
             <br/>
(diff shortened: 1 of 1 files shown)


templates/scripts/floaters/credits.php
```diff
@@ -50,16 +48,22 @@
...
-             <a href="http://www.phpconcept.net" rel="nofollow" target="_blank">PclZip</a>,
-             <a href="https://getcomposer.org/" target="_blank">Composer</a>,
-             <a href="http://mysql.com/" target="_blank">MySQL</a>,
-             <a href="http://php.net/" target="_blank">PHP</a>
+                <a href="http://framework.zend.com/" target="_blank">Zend Framework</a>,
+                <a href="http://jquery.com/" target="_blank">jQuery</a>,
+                <a href="http://phpjs.org/" target="_blank">phpjs</a>,
+                <a href="https://github.com/js-cookie/js-cookie" target="_blank">js-cookie</a>,
+                <a href="https://github.com/malsup/form" target="_blank">jQuery Form Plugin</a>,
+                <a href="https://github.com/SamWM/jQuery-Plugins/tree/master/newsticker" target="_blank">jQuery newsticker</a>,
```
kevinpapst, line 56: Do we use that? Wow, didn't know that.
And jQuery selectboxes as well?

simonschaufi, line 56: if not, I wouldn't have added it ;)
the newsticker is used in sliding through the customer, project and activity next to the buzzer when it is running.
The selectboxes as well, yes.

kevinpapst, line 56: Ah, thanks for the info, Kimai is a mystery, even after so many years working with it^^

PR #671, opened 2016-04-10
added missing translation

kevinpapst:
- added missing translations in german for user preference floater


simonschaufi:
merge also into develop?

kevinpapst:
Yes, we could merge it directly into develop. Not sure what the best way is.
I guess I wuold merge merge master completely into develop later (when we did the release?!).

simonschaufi:
from what i found out of gitflow documentation it should get merged into both branches right away so that we are always up to date on develop

Issue #403, opened 2014-07-11
Timestamp Fields Swapped in Dialog


martinvahi:
![The comment](http://longterm.softf1.com/2014/flaw_reports/kimai/kimai_v_0_9_3_rc1_dialog_swap_timestamp_t1.jpeg)

[Src downloaded 2014_07_10 from](https://github.com/kimai/kimai/releases/download/v0.9.3-rc.1/kimai_0.9.3.rc1.zip)

simonschaufi:
closed as duplicate to #507

Issue #266, opened 2013-09-29
Handle wrong server times better


ServiusHack:
Forum post: http://forum.kimai.org/index.php?topic=2310.0

It could be that the server time is off. In that case some problems can arise. For instance, when starting a recording the client time is used, when stopping it the (wrong) server time is used. Maybe there are even more cases.

Since Kimai's most important part is to record time we should do our best to handle cases when the server time and client time differ.

simonschaufi:
closed as duplicate to #507

HenryNe:
Oh oh, that is not good. I have simple tested:
1. Local time 14:00 UTC, Server time 13:00 UTC
2. Press start and stop
3. Recorded: In=14:00 Out=13:00 h'm=-1:00 rate=-1,00

kevinpapst:
@andreykl - if you want be able to choose whether client or server time is used, please raise a new issue. For the next release, we changed to using server time completely, as this is the easiest way to prevent a couple of possible errors.

PR #680, opened 2016-04-22
Use server time (resolves #507)

### LINKED ISSUE #507: Accurate time tracking

HenryNe:
FIXES #507, #266, #403

- Adjust offset for ticktac (Buzzer)
- "offset" is timediff between server and local, only set ones on page load
- "startsec" is server time
- "now" is local time

Reason for this pull request:

How to test:
4. After this patch recorded: In=13:00 Out=13:00 h'm=0:00 rate=0,0'


kevinpapst:
Hey @HenryNe - thanks again for contributing!
Please make sure that you have a clean commit history and learn how to keep your local git repo in sync with upstream. This should not be done by merging the remote, but by using a fast-forward or rebase.
Please do a squash commit, followed by a force push.

HenryNe:
Git was very restive. Nothing of the squash was perfectly.
I checked out a fresh source from your develop and added my changes again.
Please use my new clean PR #681.

kevinpapst:
It gets easier by the time^^

Did you see https://github.com/kimai/kimai/blob/master/.github/CONTRIBUTING.md#keep-your-fork-in-sync-with-original-repository
Its just as simple as that (and you can replace master in the example calls with develop as well).
If you squash an already pushed branch, you have to "force push" - otherwise git is complaining loudly ;-)

simonschaufi:
if you use phpstorm, rebase is actually super easy. you can do it with a few clicks without entering any git command! do you use phpstorm?

HenryNe:
Rebasing was not possible, because there was many other commits from me in my _develop_.
I have clean fetched from _upstream/develop_ now, and as long I do never commit something into my _develop_ I can merge and rebase without problems in future.

simonschaufi:
don't rebase on YOUR origin/develop, rebase on upstream/develop! that should always work

kevinpapst:
Or pull origin/develop into your develop (which you should not change anyway) and than rebase onto that

HenryNe:
> Or pull origin/develop into your develop (which you should not change anyway) and than rebase onto 

This is what I'm doing now.

PR #681, opened 2016-04-22
Use server time in startRecord (resolves #507), cleanup

### LINKED ISSUE #507: Accurate time tracking

HenryNe:
FIXES #507, #266, #403

- Adjust offset for ticktac (Buzzer)
- "offset" is timediff between server and local, only set ones on page load
- "startsec" is server time
- "now" is local time

Reason for this pull request:

How to test:
4. After this patch recorded: In=13:00 Out=13:00 h'm=0:00 rate=0,00

### DIFF
--- a/extensions/ki_timesheets/js/ts_func.js
+++ b/extensions/ki_timesheets/js/ts_func.js
@@ -278,8 +278,7 @@ function ts_ext_recordAgain(project,activity,id) {
     $('#timeSheetEntry'+id+'>td>a.recordAgain>img').attr("src","../skins/"+skin+"/grfx/loading13.gif");
     hour=0;min=0;sec=0;
     now = Math.floor(((new Date()).getTime())/1000);
-    offset = now;
-    startsec = 0;
+    startsec = now - offset;
     show_stopwatch();
     $('#timeSheetEntry'+id+'>td>a').removeAttr('onclick');
 
--- a/js/main.js
+++ b/js/main.js
@@ -337,11 +337,10 @@ function updateTimeframeWarning() {
 function startRecord(projectID,activityID,userID) {
     hour=0;min=0;sec=0;
     now = Math.floor(((new Date()).getTime())/1000);
-    offset = 0;
-    startsec = now;
+    startsec = now - offset;
     show_stopwatch();
     value = projectID +"|"+ activityID;
-    $.post("processor.php", { axAction: "startRecord", axValue: value, id: userID, startTime: now},
+    $.post("processor.php", { axAction: "startRecord", axValue: value, id: userID},
         function(response){
             var data = jQuery.parseJSON(response);
             currentRecording = data['id'];
@@ -480,8 +479,7 @@ function buzzer_preselect_update_ui(selector,selectedID,updateRecording) {
 // ... so just THX! ;)
 
 function ticktac() {
-    startsecoffset = startsec ? startsec : offset;
-    sek   = Math.floor((new Date()).getTime()/1000)-startsecoffset;
+    sek   = Math.floor((new Date()).getTime()/1000) - startsec - offset;
     hour  = Math.floor(sek / 3600);
     min   = Math.floor((sek-hour*3600) / 60);
     sec   = Math.floor(sek-hour*3600-min*60);
--- a/core/processor.php
+++ b/core/processor.php
@@ -144,7 +144,7 @@ switch ($axAction) {
         }
     
         $IDs = explode('|', $axValue);
-        $newID = $database->startRecorder($IDs[0], $IDs[1], $id, $_REQUEST['startTime']);
+        $newID = $database->startRecorder($IDs[0], $IDs[1], $id);
         echo json_encode(array(
           'id' =>  $newID
         ));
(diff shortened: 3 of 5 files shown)


extensions/ki_timesheets/js/ts_func.js
```diff
@@ -278,8 +278,7 @@ function ts_ext_recordAgain(project,activity,id) {
     $('#timeSheetEntry'+id+'>td>a.recordAgain>img').attr("src","../skins/"+skin+"/grfx/loading13.gif");
     hour=0;min=0;sec=0;
     now = Math.floor(((new Date()).getTime())/1000);
-    offset = now;
-    startsec = 0;
+    startsec = now - offset;
```
kevinpapst, line 281: Could you please explain this change, I can't wrap my head around it.
What does it have to do with the server time change and what is the outcome of this change?

js/main.js
```diff
@@ -480,8 +479,7 @@ function buzzer_preselect_update_ui(selector,selectedID,updateRecording) {
 // ... so just THX! ;)
 
 function ticktac() {
-    startsecoffset = startsec ? startsec : offset;
-    sek   = Math.floor((new Date()).getTime()/1000)-startsecoffset;
+    sek   = Math.floor((new Date()).getTime()/1000) - startsec - offset;
```
kevinpapst, line 482: Same as above, I get the backend changes, but I am not familiar enough with these old javascript parts. Why did you change these global variables?

kevinpapst:
Ok, this really needs to be fixed. Your solution is simple and effective and I guess we can merge it (after I understood the javascript changes).

We should be consistent and use server time for both actions (start and stop). 
What do you think @simonschaufi ? 

In the future we should still think about passing in a full ISO8601 Date object from javascript to determine the timezone of the user (or use its profile setting) and add that to the record for displaying purpose (but thats a complete different topic).

extensions/ki_timesheets/js/ts_func.js
HenryNe, line 281: This change is for the stopwatch near Buzzer.

If you click the record again, then the buzzer will run stopwatch.
The buzzer calculated the timediff from "startsec", but ticktac needs this "startsec" as server based time.

"now" is the client (browser) time.

With the "now - offset" we calculate the real server time, based on the initial time "offset", calculated on loading this page.

On loading page it was calculated the difference between server and browser as follow:
`offset =  browserTime - serverTime`

Every place you need "serverTime", you can calculate it by changing the formula above into:
`serverTime = browserTime - offset`

Of curse the primary change to save time only at server is the modification in php `function startRecorder` (see my step 8bd9dbc)
Without this modification, the stopwatch would display wrong count.

_23.04.2016: Text spelling updated_

HenryNe: I must also say, that I not understand, why the previous variant was written so complicated. Before my modification the variable "startsec" contains a start sec relative (start with 0), if the buzzer was started from inside java. The same variable "startsec" contained an absolute time (based on 1970, I think), if the page was reloaded.

After my modification the "startsec" contains the absolute time of start, adjusted to the server base in all cases (start from buzzer, "start again" from time sheed, or start from reloading the page).
To get the counted time for buzzer, ticktac  must simple differ "startsec" from current time and respect the offset between server and local time.

js/main.js
HenryNe, line 482: I must also say, that I not understand, why the previous variant was written so complicated.
Before my modification the variable "startsec" contains a start sec relative (start with 0), if the stopwatch was started from inside javascript. The same variable "startsec" contained an absolute time (based on 1970, I think), if the page was reloaded.

After my modification the "startsec" contains the absolute time of start, adjusted to the server base in all cases (start from buzzer, "start again" from time sheet, or start from reloading the page).
To get the counted time for stopwatch, ticktac must simple differ "startsec" from current time and respect the offset between server and local time.

_23.04.2016: updated spelling_

HenryNe, line 482: The formula in line 482 would be better readable in this form:

``` javascript
now_Local_based = Math.floor((new Date()).getTime()/1000);  /* current browser timestamp */
now_Server_based = now_Local_based - offset;       /* timestamp adjusted to server time */
start_Server_based = startsec;                     /* start time from server */
sek = now_Server_based - start_Server_based;       /* timediff from same base time */
```

extensions/ki_timesheets/js/ts_func.js
simonschaufi: declare new local variable with `var`

js/main.js
simonschaufi, line 482: can we extract this calculation into a function? then you can use better internal variable names so that other people also understand what's going on.
I would expect something like this:

``` javascript
sek = getTimeDiff(...);
```

HenryNe, line 482: I want not to change to many things. It is a simple timediff calculation from global variables with different time bases.
I would prefer a comment for the calculation like this:

```
/* Timediff calculation:
   "sek" = ("now" as local time - "offset" to server) - "startsec" from server */
```

simonschaufi, line 482: I would prefer doing it. It makes the code much more readable and easier to test

kevinpapst, line 482: If you understand the code, please go ahead and add these comments (we have a lot of old code from various developer, that we have never touched before, so we are happy about any comment).

HenryNe:
This pull request was splitted into two requests #684 and #685 now.

PR #683, opened 2016-04-23
ticktac: Use modulo and trunc for integer division

HenryNe:
code optimization

- Use step by step modulo operator for integer division
- Use `trunc` to get the integral part from number 

Reason for this pull request:
- Not need to check the case, if minute or hour is 60 or grater 59
- `floor` rounds down, `trunc` strips all from right side of comma.
- `floor` makes a different for negative numbers, that of curse we should never have here.
  Example `Math.floor(-1.2) => 2`  and `Math.trunc(-1.2) => 1` [refer mozilla.org](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc)

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -473,18 +473,11 @@ function ticktac() {
 
     // Split total seconds from start time to current time into
     // separate variables for viewing hours:minutes:seconds
-    var hours   = Math.floor(total_seconds / 3600);
-    var minutes = Math.floor((total_seconds - hours * 3600) / 60);
-    var seconds = Math.floor(total_seconds - hours * 3600 - minutes * 60);
+    var total_minutes = Math.trunc(total_seconds / 60);
 
-    if (seconds == 60) {
-        seconds = 0;
-        minutes++;
-    }
-    if (minutes > 59) {
-        minutes = 0;
-        hours++;
-    }
+    var hours = Math.trunc(total_minutes / 60);
+    var minutes = total_minutes % 60;
+    var seconds = total_seconds % 60;
 
     $("#s").html(prependZeroIfNeeded(seconds));
     $("#m").html(prependZeroIfNeeded(minutes));


js/main.js
```diff
@@ -482,12 +482,11 @@ function buzzer_preselect_update_ui(selector,selectedID,updateRecording) {
 function ticktac() {
     startsecoffset = startsec ? startsec : offset;
     sek   = Math.floor((new Date()).getTime()/1000)-startsecoffset;
-    hour  = Math.floor(sek / 3600);
-    min   = Math.floor((sek-hour*3600) / 60);
-    sec   = Math.floor(sek-hour*3600-min*60);
+    sec   = sek % 60;
+    rem   = Math.trunc(sek / 60);
```
simonschaufi: - declare local variable with `var`
- no abreviations in variable names

kevinpapst:
I am not a JS expert, so I a might be wrong, but you are now declaring all the variables in a local scope, where as before they where used from the global scope (defined at https://github.com/kimai/kimai/blob/master/templates/scripts/core/main.php#L82 and following lines).

So your changes mess around heavily with these global variables, doesn't it?

js/main.js
```diff
@@ -481,13 +481,12 @@ function buzzer_preselect_update_ui(selector,selectedID,updateRecording) {
 
 function ticktac() {
     startsecoffset = startsec ? startsec : offset;
-    sek   = Math.floor((new Date()).getTime()/1000)-startsecoffset;
-    hour  = Math.floor(sek / 3600);
-    min   = Math.floor((sek-hour*3600) / 60);
-    sec   = Math.floor(sek-hour*3600-min*60);
+    var sek   = Math.floor((new Date()).getTime()/1000)-startsecoffset;
+    var sec   = sek % 60;
+    var tmp   = Math.trunc(sek / 60);
```
kevinpapst: Any reason for having that variable at all (other than for readability)?

HenryNe: Have not seen any (read) usage of these variables as in ticktac. More thinking about it, I would remove all others globals hour, min, sec.
Not need to set hour=0, min=0, sec=0 _and_ startsec. They are overwritten before used in ticktac. ticktac depends only on startsec and offset.
Should I send separate PR for deleting all `hour=0;min=0;sec=0;`, or include here?
In the ticktac are more local variables, that never used outside: htmp, mtmp, stmp, titleclock.

HenryNe: You can see removing the global vars in PR #689.

kevinpapst: You are really opening and closing to many issues ;-) Why not include it here?
This PR is obsolete then in favor of #689 ?

HenryNe: Not, this PR is not obsolete then. #689 is additional.
This change on _modulo and trunc_ have nothing to do with the cleanup global variables.

Sorry :-|, I followed CONTRIBUTING.md :

> You can restructure the code, but that should be done in an independent branch
> Make sure the Pull request has a decent size

HenryNe:
For your question:

> Any reason for having that variable at all (other than for readability)?

Need not all, but better readable.
sec, min and hour need of curse.
sek and tmp can be the same, for example:
```javascript
var tmp   = Math.floor((new Date()).getTime()/1000) - startsec - offset;
var sec   = tmp % 60;
    tmp   = Math.trunc(tmp / 60);
var min   = tmp % 60;
var hour  = Math.trunc(tmp / 60);
```
My favorite is in the PR. I'm not love using same variable for different things.

js/main.js
HenryNe: For your question:

> Any reason for having that variable at all (other than for readability)?

Need not all, but better readable.
sec, min and hour need of curse.
sek and tmp can be the same, for example:

``` javascript
var tmp   = Math.floor((new Date()).getTime()/1000) - startsec - offset;
var sec   = tmp % 60;
    tmp   = Math.trunc(tmp / 60);
var min   = tmp % 60;
var hour  = Math.trunc(tmp / 60);
```

My favorite is in the PR. I'm not love using same variable for different things.

simonschaufi: > Sorry :-|, I followed CONTRIBUTING.md

we mean by that that one does not change several independent things (not talking about lines but about topics)

HenryNe:
``` javascript
var ticktac_seconds = Math.floor((new Date()).getTime()/1000) - startsec - offset;
var ticktac_minutes = Math.trunc(ticktac_seconds / 60);
var hour  = Math.trunc(ticktac_minutes / 60);
var min   = ticktac_minutes % 60;
var sec   = ticktac_seconds % 60;
```

is that better?

simonschaufi: make it clear what is really stored, especially in the first 2 lines. You can make the variable even longer so that everybody understands whats the true meaning of it ;)

``` javascript
var ticktac_seconds = Math.floor((new Date()).getTime() / 1000) - startsec - offset;
var ticktac_minutes = Math.trunc(ticktac_seconds / 60);
var hours = Math.trunc(ticktac_minutes / 60);
var minutes = ticktac_minutes % 60;
var seconds = ticktac_seconds % 60;
```

HenryNe: Done. Say, If I should squash all commits into single one.

HenryNe:
Have rebased this pull requeset to have only one clean commit.
StyleCI is broken, but not in my changed files.

simonschaufi:
As this is a fundamental change, I would like to ask what is not working as expected. Please give some examples how it will affect time tracking

kevinpapst:
I like it, just makes the code look cleaner and more professional ;-)

HenryNe:
@simonschaufi The old code is working, because the two "if ... > 59" fixes the problems. This is a "code optimization".

kevinpapst:
Conflicts :(

HenryNe:
Merged from develop. Conflict solved.

HenryNe:
it's outdated now

Issue #474, opened 2015-03-17
time error and once enter comment, project's "in" time and "out" time will change


chongwj:
Hi All,
Now i'm facing a serious problem on timesheet system.

First problem is time error : 
cannot edit that time, h'm always show an negative hours and minutes
![timesheet1](https://cloud.githubusercontent.com/assets/11517689/6685097/d7c0476e-ccd1-11e4-95ed-928d2a75bd5f.png)
...please refer to picture (timesheet1.png) . thanks

No.2 , once i press edit buttom on indivual task, project's "in" and "out" will change.
Once i press the edit key and did not change edit entry panel anything then i click OK.The "In" and"Out" will auto change. Please refer to picture( timesheet2.png and timesheet3.png)
THANKS AND REGARDS
![timesheet2](https://cloud.githubusercontent.com/assets/11517689/6685284/5dfce48a-ccd3-11e4-83a0-8ebae9e063e9.png) 
![timesheet3](https://cloud.githubusercontent.com/assets/11517689/6685290/69fa5c2c-ccd3-11e4-8b64-7311cf1b96a4.png)

.

ServiusHack:
I need some more information to help you with these problems.

What version of Kimai are you running?

What happens if you try to edit the entry with a negative duration? It should still open the edit dialog and allow you to change the start and end time.

Regarding your second problem: What is shown when you reload the page?

Are there any errors in temporary/logfile.txt or your PHP error log file?

chongwj:
Hi ,
Glad to have your reply. :)
First problem
1,KIMAI Version :-
* 0.9.3.1384*

*2,If i try to edit the entry with negative duration...system will show
"The start time must be before the end time"*


*Second problem*

*1,What is shown when you reload the page :- No different2,*
*Are there any errors in temporary/logfile.txt or your PHP error log file?
:- think NO*
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
>
> *[24.03.2015 13:10:06] login: admin as user[24.03.2015 13:10:07] Global
> role 1 gave true for ki_budget-access.[24.03.2015 13:10:07] Global role 1
[...]

PR #684, opened 2016-04-23
Use server time in startRecord (resolves #507)

### LINKED ISSUE #507: Accurate time tracking

HenryNe:
FIXES #507, #266, #403, #474

Reason for this pull request:

How to test:
4. After this patch recorded: In=13:00 Out=13:00 h'm=0:00 rate=0,00

_This is a subset of pull request #681. This request does not fix similar problem on stopwatch, which was moved into #685._

### DIFF
--- a/core/processor.php
+++ b/core/processor.php
@@ -144,7 +144,7 @@ switch ($axAction) {
         }
     
         $IDs = explode('|', $axValue);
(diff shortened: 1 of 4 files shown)


simonschaufi:
@kevinpapst include this in release?

kevinpapst:
Thanks @HenryNe 👍 this one fixes some very old complaints - lets hope everyone likes it!

PR #687, opened 2016-04-24
[WIP] Refactor permissions

kevinpapst:
**THIS IS WORK IN PROGRESS!**

I opened a Pull request to let you know that I am working on it currently and that we should not change anything related to permissions until I am done here.

Fixes https://github.com/kimai/roadmap/issues/5

- completely restructure the permission system by changing the underlying database structure (which is broken by design, as it stops us from having dynamic included extensions)


simonschaufi:
would be really nice if we could get this one done.

kevinpapst:
will not be implemented

PR #690, opened 2016-04-24
startRecord without ts_ext_reload (resolves #493)

### LINKED ISSUE #493: Buzzer gets inactive again right after starting it
Through debugging of the javascript i found out that (on one server only) the requests and responses vary between different servers which result in different results.

1. Request: axAction:"setTimeframe"
2. Request: axAction:"startRecord" 
3. Request: axAction:"reload_timeSheet" with response content: updateRecordStatus(false);
4. Request: axAction:"reload_timespan" with response content: Zeitspanne:  <b>xx.xx.2015 - xx.xx.2015</b>
5. Request: axAction:"reload_timeSheet" with response content: updateRecordStatus(1286, 1434502313,...)

The problem lies here at the 3rd request. On my localhost i get the response like this:
updateRecordStatus(1276, 1434502997,...

It is reproducable but I don't know where exactly lies the problem as the files AND the db are identical on both systems.

I just know that here the $kga['customer'] variable is not set on one system and the other one it is set: https://github.com/kimai/kimai/blob/master/core/extensions/ki_timesheets/init.php#L60

HenryNe:
FIXES #493

- Do not call extra ts_ext_reload

Reason for this pull request:
- ts_ext_reload is called via setTimeframe, via hook_timeframe_changed and  timesheet_extension_timeframe_changed
- ts_ext_reload should not call, if timesheet is not active. This will be perfect handled in timesheet_extension_timeframe_changed

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -346,7 +346,6 @@ function startRecord(projectID,activityID,userID) {
             var data = jQuery.parseJSON(response);
             currentRecording = data['id'];
             $('#buzzer').removeClass('disabled');
-            ts_ext_reload();
         }
     );
 }


simonschaufi:
yeah, thank you very very much! this really solved my problem!

HenryNe:
Thank to ServiusHack ;-)

There was written a good analytics:

https://github.com/kimai/kimai/issues/493#issuecomment-118631699

PR #699, opened 2016-05-05
End time isn't updated after entering in format HH.mm

### LINKED ISSUE #695: TS: End time isn't updated after entering in format HH.mm
#### Expected behaviour
If you enter the start time in the time sheet in the format HH.mm.ss or HH.mm (e.g. 10.00) it will change the format after losing focus to HH:mm:ss. Which is good because nothing else is accepted when submitting the form. This should work for the end time as well.

#### Actual behaviour
But it doesn't work for the end time. When the input field loses focus/a change is registered the entered time string stays in its format (e.g. 10.00) instead of reformatting to the accepted form HH:mm.

Add a new time to the timesheet. Select a project, an activity, enter a description, leave the date as is, enter for the start time 10.00 and for the end time 11.00, click on submit.

The reason the point notation is handy: You can use your numblock on the keyboard.

#### Kimai version
Just checked with the latest master.

#### How to solve it
In the file ./extensions/ki_timesheets/js/ts_func.js
Line 475, function ts_durationToTime
I added the new function ts_formatTime and ts_formatDate and changed ts_durationToTime to use them and also to put the value for \#end_time and \#end_day which wasn't done before.
```javascript
function ts_formatTime(value) {
    var H = value.getHours();
    var i = value.getMinutes();
    var s = value.getSeconds();

    if (H<10) H = "0"+H;
    if (i<10) i = "0"+i;
    if (s<10) s = "0"+s;
    
    return H + ":" + i + ":" + s;
}

function ts_formatDate(value) {
    var d = value.getDate();
    var m = value.getMonth() + 1;
    var y = value.getFullYear();
    
    if (d<10) d = "0"+d;
    if (m<10) m = "0"+m;
    
    return d + "." + m + "." + y;
}

function ts_durationToTime() {
    end = ts_getEndDate();
    durationArray=$("#duration").val().split(/:|\./);
    if(end!=null && durationArray.length > 0 && durationArray.length < 4) {
        secs = durationArray[0]*3600;
        if(durationArray.length > 1)
            secs += (durationArray[1]*60);
        if(durationArray.length > 2)
            secs += parseInt(durationArray[2]);
        begin = new Date();
        begin.setTime(end.getTime()-(secs*1000));

        $("#start_time").val(ts_formatTime(begin));
        $("#end_time").val(ts_formatTime(end));
        $("#start_day").val(ts_formatDate(begin));
        $("#end_day").val(ts_formatDate(end));
    }
}
```

Dravere:
FIXES #695 

- Extracted formatting of date and time to be reused
- Formatted not only the start time and date but also the end time and date in ts_durationToTime, see #695 for further explanation

Reason for this pull request:
- It was requested in the issue #695

### DIFF
--- a/extensions/ki_timesheets/js/ts_func.js
+++ b/extensions/ki_timesheets/js/ts_func.js
@@ -31,6 +31,21 @@ function ts_ext_onload() {
 }
 
 /**
+ * Formats a date object to be used in the time input field.
+ */
+function ts_formatTime(value) {
+    var H = value.getHours();
+    var i = value.getMinutes();
+    var s = value.getSeconds();
+
+    if (H<10) H = "0"+H;
+    if (i<10) i = "0"+i;
+    if (s<10) s = "0"+s;
+
+    return H + ":" + i + ":" + s;
+}
+
+/**
  * Update the dimension variables to reflect new height and width.
  */
 function ts_ext_get_dimensions() {
@@ -484,16 +499,7 @@ function ts_durationToTime() {
         begin = new Date();
         begin.setTime(end.getTime()-(secs*1000));
 
-
-        var H = begin.getHours();
-        var i = begin.getMinutes();
-        var s = begin.getSeconds();
-
-        if (H<10) H = "0"+H;
-        if (i<10) i = "0"+i;
-        if (s<10) s = "0"+s;
-
-        $("#start_time").val(H + ":" + i + ":" + s);
+        $("#start_time").val(ts_formatTime(begin));
 
         var d = begin.getDate();
         var m = begin.getMonth() + 1;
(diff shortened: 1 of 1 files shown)


extensions/ki_timesheets/js/ts_func.js
```diff
@@ -31,6 +31,35 @@ function ts_ext_onload() {
...
 
 /**
+ * Formats a date object to be used in the time input field.
+ */
+function ts_formatTime(value) {
+    var H = value.getHours();
+    var i = value.getMinutes();
+    var s = value.getSeconds();
+
+    if (H<10) H = "0"+H;
```
simonschaufi: Please follow PSR Code style here and reformat the code like this:

``` javascript
if (H < 10) {
    H = '0' + H;
}
```

please also use lowercase and speaking (hour) variable names.

Dravere:
You're aware that you requested style adjustments for code that simply was extracted from yours? But sure, here we go. Adjusted it for PSR.

PR #715, opened 2016-06-16
Updated italian translation

massiws:
Added some translation strings.


simonschaufi:
Thank you very much for your contribution!

PR #716, opened 2016-06-19
Bugfix: Check if user exists

### LINKED ISSUE #701: Adding a user that already exists - empty error message
#### Expected behaviour
To be able to add a user with a blank email address

#### Actual behaviour
'Edit User 'screen prompts "Password not set" and a yellow exclamation triangle appears for that user's 'Status' in the summary screen

Add a user with no email address set

#### Kimai version
0.9.3.1387

I expect to be able to add users with no email address. I also expect that the option to 'Email this user' to be greyed-out if there's no valid email address but, of course, I haven't been able to test that :-)

simonschaufi:
When creating a new user, a check is missing if the user exists already.

Resolves #701

### DIFF
--- a/extensions/ki_adminpanel/processor.php
+++ b/extensions/ki_adminpanel/processor.php
@@ -34,12 +34,20 @@ switch ($axAction)
         $groupsWithAddPermission = array();
         foreach ($kga['user']['groups'] as $group) {
             $membershipRoleID = $database->user_get_membership_role($kga['user']['userID'], $group);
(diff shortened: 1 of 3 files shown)


kevinpapst:
Looks good! But why to master and not develop?

simonschaufi:
I thought that bugfixes go straight to master and features into develop (at least thats the gitflow way)

kevinpapst:
I wouldn't neccessarily say that this is a bug. But well, we have to merge master to develop later anyway...

kevinpapst:
Hey, please use branches from your fork :)

PR #717, opened 2016-06-19
	Use glob to find languages

simonschaufi:
If we decide to drop the PR, we should instead find a better solution for files iteration. My suggestion is to use [glob](http://php.net/glob).

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -63,13 +63,13 @@ switch ($axAction) {
         $skins = array();
         $langs = array();
 
-        $allSkins = glob(__DIR__ . "/../skins/*", GLOB_ONLYDIR);
+        $allSkins = glob(__DIR__ . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR . 'skins' . DIRECTORY_SEPARATOR . '*', GLOB_ONLYDIR);
         foreach ($allSkins as $skin) {
             $name = basename($skin);
             $skins[$name] = $name;
         }
 
-        foreach (Kimai_Translations::langs() as $lang) {
+        foreach (Kimai_Translations::languages() as $lang) {
             $langs[$lang] = $lang;
         }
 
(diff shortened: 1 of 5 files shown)


kevinpapst:
The new class seems to be perfect testable ...
And yes, using glob seems more elegant ;-) but both version are fine

simonschaufi:
ok then let's not introduce more code that's not needed. I copied the class anyway from another framework ;)

kevinpapst:
Close then?

kevinpapst:
Ups, wrong button - you can reopen if desired

simonschaufi:
undoing a pr is only possible with a revert commit.

kevinpapst:
Why a milestone if nothing happened?

simonschaufi:
since it's merged into master, it will be in the 1.0 release.

kevinpapst:
Nope, it wasn't merged.

kevinpapst:
But now we need some Unit Tests!

simonschaufi:
I will take care of it

kevinpapst:
Might be a little picky, but Kimai_Utility_StringUtility sounds strange.
Why not Kimai_Utility_String or Kimai_Utility_StringHelper?

And UnitTests - then fine to merge

simonschaufi:
No compromise with the class name. The guys from TYPO3 already thought about a good name already ;)

kevinpapst:
Is there a way to include it directly from Typo3 via composer? 
This seems to be a 1:1 copy of their code and we cannot simply copy&paste their code into our project.

simonschaufi:
You can only include whole projects, not single files via composer.

That is what the GNU License says:

https://github.com/TYPO3/TYPO3.CMS/blob/master/LICENSE.txt#L85-L91

So the only thing that is missing is the mention of TYPO3

kevinpapst:
Yeah, for some (sorry) stupid String functions... I am not convinced.

simonschaufi:
Using glob makes more sense...

kevinpapst:
Actually, if this is the only place we use this function: I totally support that!
What about updating that PR to use glob and remove the new class? And this is fine for merging...

PR #723, opened 2016-06-26
Fix remote api

kevinpapst:
FIXES https://github.com/kimai/kimai-mobile/issues/15

- do not return hidden entities via JSON/SOAP


simonschaufi:
Either provide a unit test or a way to test this manually

kevinpapst:
You could use the kimai-mobile version and check beforee and after. Projects and such should not include hidden entries any longer

simonschaufi:
I'm just trying to write some unit tests but the API class is not so easily testable.

kevinpapst:
Uh, nice! I know, like all the old classes ... thats why I skipped it ^^

PR #728, opened 2016-07-03
Release changes 1.0.1

kevinpapst:
Fixes after the 1.0.0 release.
- Version adjustments
- License header

### DIFF
--- a/includes/version.php
+++ b/includes/version.php
@@ -1,4 +1,4 @@
 <?php
-$kga['version'] = '1.0.0';
-$kga['revision'] = 1387; // database revision number (incremented whenever the database changes)
-$kga['status'] = 'dev'; // leave blank if stable
+$kga['version'] = '1.0.1';
+$kga['revision'] = 1388; // database revision number (incremented whenever the database changes)
+$kga['status'] = ''; // leave blank if stable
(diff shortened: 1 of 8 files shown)


includes/version.php
```diff
@@ -1,4 +1,4 @@
 <?php
-$kga['version'] = '1.0.0';
-$kga['revision'] = 1387; // database revision number (incremented whenever the database changes)
-$kga['status'] = 'dev'; // leave blank if stable
+$kga['version'] = '1.0.1';
+$kga['revision'] = 1388; // database revision number (incremented whenever the database changes)
+$kga['status'] = ''; // leave blank if stable
```
simonschaufi, line 4: could you explain why you raised this to 1388 (not existing) and empty status (only for release, should be something else for current development)?

kevinpapst, line 4: Because the updater is stuck if ONLY the version is increased. So everyone who used the 1377 with 0.9.3 before, could not get around the updater - unless he would change the database value manually.
This seems to be a bug in the updater (the last lines which update the revision).
You can easily test this, by decreasing the version number only in your installation.

For "develop" branch we can change the value back to dev again.

simonschaufi, line 4: I would go for dev in master too as any new change will not be the same like the last release (for bug reports and so on). It is clear that people are obviously using the github version and not the release.

kevinpapst, line 4: Sure, after the release we can change it back.

PR #731, opened 2016-07-05
Fix cache directory

### LINKED ISSUE #729: Cannot add or edit timesheet entry
#### Expected behaviour
I click edit on an existing timesheet entry. The popup dialog opens, i can edit values and clicking OK will submit/save the timesheet data.
Directory permissions are alright. 777 for 'temporary'.

#### Actual behaviour
Clicking OK results in an XHR request to the URL https://HOSTNAME/extensions/ki_timesheets/processor.php. Response code is 200. But the edit dialog doesn't close and also the timesheet data is not saved.

Click edit on a timesheet entry. Edit duration. Click OK.

#### Kimai version
Kimai 1.0.1.1388

[05.07.2016 00:06:39] Uncaught exception: Could not determine temp directory, please specify a cache_dir manually 

#### Additional information
It seems this bug existed before: https://github.com/kimai/kimai/issues/483.

I applied the proposed hack in File.php (Zend_Cache Backend) and hardcoded the absolute path to the temporary directory. This solves the problem. 

Shouldn't the installer somehow configure the cache-directory?

kevinpapst:
FIXES #729

- adds a global Zend_Cache object and injects it into Zend components, which are indirectly used

Reason for this pull request:
- some servers have no working temp/ directory configured 
- Kimai has a temporary/ directory, so we should use it

### DIFF
--- a/includes/basics.php
+++ b/includes/basics.php
@@ -83,7 +83,7 @@ if (!empty($vars)) {
 }
 
 $tmpDir = WEBROOT . 'temporary/';
-if (!file_exists($tmpDir) || !is_dir($tmpDir) ||!is_writable($tmpDir)) {
+if (!file_exists($tmpDir) || !is_dir($tmpDir) || !is_writable($tmpDir)) {
     die('Kimai needs write permissions for: temporary/');
 }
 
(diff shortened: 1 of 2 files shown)


simonschaufi:
Will check it on my own host where I had to do this patch.

PR #732, opened 2016-07-10
Refactoring of forgot password

### LINKED ISSUE #713: "Forgot My Password" email system not working.
#### Expected behaviour
Receive an email regarding password change.

#### Actual behaviour
Nothing received. 
Everytime.

#### Kimai version
Latest

Where is the code to prompt the email to be sent?

simonschaufi:
FIXES #713

### DIFF
--- a/libraries/Kimai/Auth/Kimai.php
+++ b/libraries/Kimai/Auth/Kimai.php
@@ -49,31 +49,57 @@ class Kimai_Auth_Kimai extends Kimai_Auth_Abstract
 
     /**
      * @param string $name
-     * @return mixed
+     * @return string
+     * @throws \Zend_Mail_Exception
      */
     public function forgotPassword($name)
     {
         $kga = $this->getKga();
         $database = $this->getDatabase();
 
-        $id = $database->user_name2id($name);
+        $is_customer = $database->is_customer_name($name);
 
-        $user = $database->user_get_data($id);
+        $mail = new Zend_Mail('utf-8');
+        $mail->setFrom($kga['conf']['adminmail'], 'Kimai - Open Source Time Tracking');
+        $mail->setSubject($kga['lang']['passwordReset']['mailSubject']);
+        
+        $transport = new Zend_Mail_Transport_Sendmail();
+
         $passwordResetHash = str_shuffle(MD5(microtime()));
 
-        $database->user_edit($id, array('passwordResetHash' => $passwordResetHash));
+        if ($is_customer) {
+            $customerId = $database->customer_nameToID($name);
+            $customer = $database->customer_get_data($customerId);
+
+            $database->customer_edit($customerId, array('passwordResetHash' => $passwordResetHash));
+
+            $mail->addTo($customer['mail']);
+        } else {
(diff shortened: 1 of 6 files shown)


libraries/Kimai/Auth/Kimai.php
```diff
@@ -48,52 +48,109 @@ public function authenticate($username, $password, &$userId)
...
     public function forgotPassword($name)
     {
         $kga = $this->getKga();
         $database = $this->getDatabase();
 
-        $id = $database->user_name2id($name);
+        $is_customer = $database->is_customer_name($name);
+
+        $mail = new Zend_Mail('utf-8');
+        $mail->setFrom($kga['conf']['adminmail'], 'Kimai - Open Source Time Tracking');
```
kevinpapst, line 63: This could be a translation key or even an admin setting

libraries/Kimai/Auth/Kimai.php
```diff
@@ -48,52 +48,109 @@ public function authenticate($username, $password, &$userId)
...
+            
+            $database->user_edit($userId, array('passwordResetHash' => $passwordResetHash));
+
+            $mail->addTo($user['mail']);
+        }
+        
+        Kimai_Logger::logfile('password reset: ' . $name . ($is_customer ? ' as customer' : ' as user'));
 
         $ssl = !empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off';
         $url = ($ssl ? 'https://' : 'http://') . $_SERVER['SERVER_NAME'] . dirname($_SERVER['SCRIPT_NAME']) . '/forgotPassword.php?name=' . urlencode($name) . '&key=' . $passwordResetHash;
```
kevinpapst, line 89: I know you didn't change that line, but does that work when Kimai is installed in a subdirectory?

libraries/Kimai/Auth/Kimai.php
simonschaufi: sure, go ahead and make a PR ;)

libraries/Kimai/Auth/Kimai.php
simonschaufi, line 89: `dirname($_SERVER['SCRIPT_NAME'])` should do it but it fails on windows since on win it contains backslashes. There should be a common utility class that handles paths. I'm trying to "steal" a lot from here like: https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Classes/Utility/PathUtility.php

libraries/Kimai/Auth/Kimai.php
simonschaufi, line 63: to make it even better, there should be a config setting which type of transport you want and if smtp is set, all authorization settings should be settable.

kevinpapst:
Your PR is hard to read, if you change so many things that are not logic.
Otherwise: Looks good!

simonschaufi:
yeah sorry, I mixed a lot of code changes with style changes...

simonschaufi:
TODO:
- provide Utility Class for URL and file paths
- Admin settings for email configuration (smtp..., Sender...)

PR #733, opened 2016-07-10
Rounding total price goes wrong

### LINKED ISSUE #480: Rounding total price goes wrong
I found a bug with rounding total prices.
Example:
- Price: 15.00
- Duration (h:m): 3:28

3:28 converts to 3+1/60x28 = 3.46666... and the report shows 3.47 hours.
The total price should then be 3.47x15.00 = 52.05. But it takes the exact value of 3+1/60x28 and then calculates the total price, which gives 52.00.

The result is not correct. It would be nice to have an option in the Kimai settings where you can set a rounding preference. The already existing option 'Use exact sums instead of adding the shown rounded values.' does not work for this. It tried to enable and disable the option, but it gives me always 52.00.

simonschaufi:
FIXES #480

- Round duration and wage

Reason for this pull request:
- Wrong calculation

Needs testing!!!

### DIFF
--- a/extensions/ki_export/private_func.php
+++ b/extensions/ki_export/private_func.php
@@ -126,7 +126,7 @@ function export_get_data($start, $end, $users = null, $customers = null, $projec
                 $arr['time_out'] = $timeSheetEntries[$timeSheetEntries_index]['end'];
                 $arr['duration'] = $timeSheetEntries[$timeSheetEntries_index]['duration'];
                 $arr['formattedDuration'] = $timeSheetEntries[$timeSheetEntries_index]['formattedDuration'];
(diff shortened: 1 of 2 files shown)


kevinpapst:
General thought: 
- Everything should be calculated the same way
- No calculating in mysql selects
- Fetch only the raw value with the select: "start-end as duration"
- Create a formatDuration function in  StringUtility class "return sprintf("%01.2f", round($duration / 3600, 2));"
- Replace all of your adjusted code pieces with the function call StringUtility::formatDuration($duration)

That would be even cleaner :)

kevinpapst (approved):

simonschaufi:
@kevinpapst did you really test that?

kevinpapst:
Hahaha, I hope you did back in the days. I checked through all my issues and PRs and saw them. 
And as I have no plans whatsoever with Kimai 1, I thought that I can approve them as well (they look ok, thats all I can say).

PR #688, opened 2016-04-24
Kga to object

kevinpapst:
Changes proposed in this pull request:
- Large refactoring of the $kga to an ArrayObject, allowing backward compatibility with the array usage
- The new object structure allows for OO programming
- Moved Authenticator to Registry (fixes kimai/roadmap#3)
- Rewritten Debug extension to not-preload data, but use JS always / allow for easier debugging the kga object / removed reload button / removed unused images
- added Kimai_Translation_Service and Kimai_Translation_Data, preventing multiple unnecessary data loads
- rewritten config loading, utilizing the new Config ArrayObjects (preventing a lot of duplicated database requests)
- removed vars.php - logic went into new Config objects
- worked on some translations
- code formattings on the fly
- fixed editLimit check in editing and creating expenses
- ... and some more tweaks, see commits

Reason for this pull request:
- Future improvements of Kimai and allowing more OO programming style

PR #734, opened 2016-07-10
Add country field for customer

### LINKED ISSUE #246: Adding country to customer properties
Currently it is not possible to enter a specific country for a customer.
This is required, when you have customers in different countries

simonschaufi:
FIXES #246


installer/install.php
```diff
@@ -175,6 +175,7 @@ function quoteForSql($input) {
   `street` varchar(255) NULL,
   `zipcode` varchar(255) NULL,
   `city` varchar(255) NULL,
+  `country` varchar(255) NULL,
```
kevinpapst: This is different from "DEFAULT NULL" in the updater below, but it might be that MySql handles it the same way.

simonschaufi: This is correct mysql syntax. You could even leave the NULL out.

kevinpapst: I know that. My point is: do not rely on implicit values. Define "default null" either twice or not at all,.

simonschaufi: for now let's just be consistent and do it "properly" in a separate commit.

updater/updater.php
```diff
@@ -1117,6 +1117,11 @@
 
 // release of kimai 1.0.0
 
+if ((int)$revisionDB < 1388) {
```
kevinpapst: don't forget to change this after merging

kevinpapst:
A general thought here: A free text field is useful for rendering in addresses/invoices. 
But for future features, a select field with a ISO code (should be available through the Zend Framework) might be better solution. Then we could probably fetch formats and currency automatically.

libraries/Kimai/Invoice/AbstractRenderer.php
```diff
@@ -124,12 +124,15 @@ public function canRender()
     
     protected function prepareCustomerArray($customer)
     {
+        global $kga;
```
simonschaufi, line 127: I wasn't sure here if this is the way to go or if there is a better solution?

kevinpapst, line 127: No, not until #688 is merged

simonschaufi: ok then let me check the invoice again and then this is ready to merge!

kevinpapst:
Cool. Haven't tried it, but looks cool!
Now the invoice renderer knows the two-letter country code, right? Does it know a translated country name as well? This must not be necessary part of this PR, but it would be cool.

simonschaufi:
Of course i render the full country name. Nobody wants iso codes in an invoice. check the lines below the "global"

ok ready for merge!

libraries/Kimai/Invoice/AbstractRenderer.php
```diff
@@ -124,12 +124,15 @@ public function canRender()
...
+        $countryName = Zend_Locale::getTranslation($customer['country'], 'country', $kga['language']);
```
kevinpapst: What happens if $customer['country'] is empty?
Should that line be included in line 135 in the isset() check?

simonschaufi: no need. it will just return nothing.

PR #741, opened 2016-07-13
Replace comma with dot for amount value

simonschaufi:
- Entering an amount of 12,12 will fail with an error
- Replace comma with dot on keyup.

Reason for this pull request:
- make it possible to enter 12,12 instead of 12.12

### DIFF
--- a/extensions/ki_expenses/js/exp_func.js
+++ b/extensions/ki_expenses/js/exp_func.js
@@ -37,14 +37,12 @@ function expense_extension_applyHoverIntent() {
 	$('#expenses').find('tr').hoverIntent({
 		sensitivity: 1,
 		interval: 500,
-		over:
-			function() {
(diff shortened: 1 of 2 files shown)


extensions/ki_expenses/js/exp_func.js
```diff
@@ -37,14 +37,12 @@ function expense_extension_applyHoverIntent() {
...
-			function() {
-				$('#expenses').find('tr').removeClass('hover');
-				$(this).addClass('hover');},
-		out:
-			function() {
-				$(this).removeClass('hover');
-			}
+		over: function() {
+			$('#expenses').find('tr').removeClass('hover');
+			$(this).addClass('hover');},
```
kevinpapst: missing break ;)

kevinpapst:
👍

PR #747, opened 2016-07-14
Fix Kimai_Config Class error

### LINKED ISSUE #742: Kimai_Config Class error in develop branch installer
_This error only occurs in the develop branch._

#### Expected behaviour
After pressing 'Install' at the final installation wizard screen, it should fade to black. After remaining black for a few seconds (at the same URL of the installer, http://localhost/kimai_master/kimai/installer/), the login page loads.

#### Actual behaviour
After pressing 'Install' at the final installation wizard screen, it fades to black before a white screen appears. The URL of this screen is http://localhost/kimai_develop/kimai/installer/install.php?accept=1&timezone=UTC

This page returns a 500 error code, and creates this error in my PHP log file:
`[Wed Jul 13 16:51:48.575376 2016] [:error] [pid 2686] [client 127.0.0.1:36178] PHP Fatal error:  Uncaught Error: Class 'Kimai_Config' not found in /srv/http/kimai_develop/kimai/includes/vars.php:50\nStack trace:\n#0 /srv/http/kimai_develop/kimai/includes/basics.php(48): require_once()\n#1 /srv/http/kimai_develop/kimai/installer/install.php(57): include('/srv/http/kimai...')\n#2 {main}\n  thrown in /srv/http/kimai_develop/kimai/includes/vars.php on line 50, referer: http://localhost/kimai_develop/kimai/installer/`

Looking in the includes folder, I can see that autoconf.php has been created, but it is an empty file.

This always occurs when pressing 'Install' at the final installation wizard screen on the latest code from the develop branch.

#### Kimai version
The latest from the master branch performs the expected behaviour (the installer says v1.0.1.1388), but the develop branch performs the actual behaviour.

Cannot be accessed as the error occurs at the installation stage.

itsmejoeeey:
FIXES #742

- added empty path to namespace for `Kimai_` in `autoload_namespaces.php`

Reason for this pull request:
- without it installation fails

### DIFF
--- a/libraries/composer/autoload_namespaces.php
+++ b/libraries/composer/autoload_namespaces.php
@@ -8,5 +8,5 @@ $baseDir = dirname($vendorDir);
 return array(
     'Zend_' => array($vendorDir . '/zendframework/zendframework1/library'),
     'PHPExcel' => array($vendorDir . '/phpoffice/phpexcel/Classes'),
-    'Kimai_' => array($vendorDir),
+    'Kimai_' => array($vendorDir . ''),
 );
--- a/libraries/composer/autoload_namespaces.php
+++ b/libraries/composer/autoload_namespaces.php
@@ -8,5 +8,5 @@ $baseDir = dirname($vendorDir);
 return array(
     'Zend_' => array($vendorDir . '/zendframework/zendframework1/library'),
     'PHPExcel' => array($vendorDir . '/phpoffice/phpexcel/Classes'),
-    'Kimai_' => array($vendorDir . ''),
+    'Kimai_' => array($vendorDir),
 );
--- a/libraries/composer/autoload_static.php
+++ b/libraries/composer/autoload_static.php
@@ -29,7 +29,7 @@ class ComposerStaticInit8c00ac27a90dfc4ac1a7acb480023ed4
         array (
             'Kimai_' => 
             array (
-                0 => 'C:\\xampp704\\htdocs\\kimai\\htdocs\\libraries',
+                0 => __DIR__ . '/../..' . '/libraries',
             ),
         ),
     );


libraries/composer/autoload_namespaces.php
```diff
@@ -8,5 +8,5 @@
 return array(
     'Zend_' => array($vendorDir . '/zendframework/zendframework1/library'),
     'PHPExcel' => array($vendorDir . '/phpoffice/phpexcel/Classes'),
-    'Kimai_' => array($vendorDir),
+    'Kimai_' => array($vendorDir . ''),
```
simonschaufi: That doesn't fix anything. The problem is here https://github.com/kimai/kimai/blob/develop/libraries/composer/autoload_static.php#L32:
`C:\\xampp704\\htdocs\\kimai\\htdocs\\libraries`

kevinpapst:
Sorry, patching these files directly would not work, as they are auto-generated.
And as Simon pointed uot, the problem is something else

simonschaufi:
@kevinpapst he should just revert that change and instead thange the other file.

kevinpapst:
Well, okay. We'll see who is faster ;-)

libraries/composer/autoload_static.php
```diff
@@ -29,7 +29,7 @@ class ComposerStaticInit8c00ac27a90dfc4ac1a7acb480023ed4
         array (
             'Kimai_' => 
             array (
-                0 => 'C:\\xampp704\\htdocs\\kimai\\htdocs\\libraries',
+                0 => __DIR__ . '/../..' . '/libraries',
```
simonschaufi, line 32: can be shortened to `__DIR__ . '/../../libraries',` but since the other paths are also in this style, I'll just merge it.

kevinpapst, line 32: Whatever you do, don't edit this file by hand, but ONLY commit the auto generated version! "composer dump-autoload"

simonschaufi, line 32: The problem is that if I execute that command, my composer writes in the absolute path `C:\\...` which is wrong. What happens when you execute it?

itsmejoeeey, line 32: When I run `composer dump-autoload`, I get exactly what I committed in `autoload_static.php`

PR #748, opened 2016-07-14
Remote api startRecord() and stopRecord() can detect the correct activity itself.

resli:
- `startRecord($apiKey, "", "")` restarts the last stopped activity,
- `stopRecord($apiKey, "")` stops the current running activity.

This is because the bde-terminal does not knows the activity to start/stop. It only wants to start the last run or stop the current one.

### DIFF
--- a/libraries/Kimai/Remote/Api.php
+++ b/libraries/Kimai/Remote/Api.php
@@ -239,6 +239,7 @@ class Kimai_Remote_Api
     /**
      * The user started the recording of an activity via the buzzer. If this method
      * is called while another recording is running the first one will be stopped.
+     * If $projectId and $activityId are empty the last activity will be restarted.
      *
      * @param string $apiKey
      * @param integer $projectId
@@ -249,6 +250,16 @@ class Kimai_Remote_Api
     {
         if (!$this->init($apiKey, 'startRecord')) {
             return $this->getAuthErrorResult();
+        }
+
+        // check for empty params
+        if ($projectId == '' && $activityId == '') {
+            // get the last used project/activity
+            $user = $this->getUser();
+            $uid = $user['userID'];
+            $ary = $this->getBackend()->get_user_config($uid);
+            $projectId = $ary['lastProject'];
+            $activityId = $ary['lastActivity'];
         }
 
         // check for valid params
@@ -277,7 +288,8 @@ class Kimai_Remote_Api
     }
 
     /**
-     * Stops the currently running recording.
+     * Stops the currently running recording. If $entryId is empty the
+     * current activity will be stopped.
      *
      * @param string $apiKey
      * @param integer $entryId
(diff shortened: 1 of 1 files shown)


libraries/Kimai/Remote/Api.php
```diff
@@ -244,13 +244,27 @@ protected function getSuccessResult(array $items, $total = 0)
...
      */
     public function startRecord($apiKey, $projectId, $activityId)
     {
         if (!$this->init($apiKey, 'startRecord')) {
             return $this->getAuthErrorResult();
         }
 
+        // check for empty params
+        if ($projectId=="" ||
+            $activityId==""
```
simonschaufi: We use PSR 2 code style. Please format your code accordingly.

resli:
I reformatted the code. I hope it's ok now.

libraries/Kimai/Remote/Api.php
```diff
@@ -244,13 +244,25 @@ protected function getSuccessResult(array $items, $total = 0)
...
+     * If $projectId or $activityId is empty the last activity will be restarted.
      */
     public function startRecord($apiKey, $projectId, $activityId)
     {
         if (!$this->init($apiKey, 'startRecord')) {
             return $this->getAuthErrorResult();
         }
 
+        // check for empty params
+        if (($projectId=="") || ($activityId=="")) {
```
simonschaufi: I wonder why styleci is not complaining but there should be spaces around the equal signs and you can also remove the inner brackets. Please use single quotes also.

libraries/Kimai/Remote/Api.php
```diff
@@ -244,13 +244,25 @@ protected function getSuccessResult(array $items, $total = 0)
      * @param integer $projectId
      * @param integer $activityId
      * @return array
+     *
+     * If $projectId or $activityId is empty the last activity will be restarted.
```
simonschaufi: This comment must be above the params

libraries/Kimai/Remote/Api.php
```diff
@@ -282,13 +294,34 @@ public function startRecord($apiKey, $projectId, $activityId)
      * @param string $apiKey
      * @param integer $entryId
      * @return boolean
+     *
+     * If $entryId is empty the current activity will be stopped.
```
simonschaufi: Dito

libraries/Kimai/Remote/Api.php
```diff
@@ -282,13 +294,34 @@ public function startRecord($apiKey, $projectId, $activityId)
...
      */
     public function stopRecord($apiKey, $entryId)
     {
         if (!$this->init($apiKey, 'stopRecord')) {
             return $this->getAuthErrorResult();
         }
 
+        // check for empty params
+        if ($entryId=="") {
```
simonschaufi: Here also

resli:
I hope everything is ok now. Sorry for the inconvenience.

libraries/Kimai/Remote/Api.php
```diff
@@ -251,6 +252,16 @@ public function startRecord($apiKey, $projectId, $activityId)
             return $this->getAuthErrorResult();
         }
 
+        // check for empty params
+        if ($projectId == '' || $activityId == '') {
```
kevinpapst: From my POV this should be changed to 
`if ($projectId == '' && $activityId == '') {`
If only one of the values is empty, this is an invalid requestby the next check for backend->isValidX()

If both are empty we could assume, that the last task should be re-started.

resli:
I agree with you and modified the commit.

kevinpapst:
Thanks @resli 👍

PR #772, opened 2016-08-12
fix https url

EagleEyeBW:
fixes loading of google fonts when kimai is used in https enviromnent


kevinpapst:
Thanks @EagleEyeBW :)

simonschaufi:
@kevinpapst that was a bit to fast. //fonts would have been the best solution

kevinpapst:
I thought about that. But as https is always supported by Google, using https always seems to be the best solution to me and not protocol independent (which is not supported by all browsers out there).

PR #775, opened 2016-08-13
fix the export summary in std_time by replacing gmdate with a new function

### LINKED ISSUE #770: wrong standard time in REPORT
#### Expected behaviour
when i take the report for the project/user it should give the correct standard/decimal time of user/project.
#### Actual behaviour
in my case decimal time is correct but standard time id only its giving completely wrong so please check the issue.

claudegel:
FIXES #770 

-  in ki_export base_export_pdf.php I<ve replaced gmdate by a new function to fix the export summary in pdf export std-time

Reason for this pull request:
- The summary was showing wrong total time because gmdate cannot handle time length bigger then 24hrs.

### DIFF
--- a/extensions/ki_export/base_export_pdf.php
+++ b/extensions/ki_export/base_export_pdf.php
@@ -96,20 +96,21 @@ class BasePDF extends TCPDF
 
         return sprintf('% 2d:%02d', $hours, $minutes);
     }
-    
+
     /**
      * Calculate time in hh:mm from duration lenght in second
      * rounded to minutes.
      *
-     * @param int $duration , duration lenght in second
+     * @param int $duration
+     *            , duration lenght in second
      * @return string time in format hh:mm
      */
     public function timeLength($duration)
     {
-        $s=$duration % 60;
-        $m=(($duration-$s) / 60) % 60;
-        $h=floor($duration / 3600);
-        return $h.":".substr("0".$m,-2)." ".$kga['lang']['export_extension']['duration_unit'];
+        $s = $duration % 60;
+        $m = (($duration - $s) / 60) % 60;
+        $h = floor($duration / 3600);
+        return $h . ":" . substr("0" . $m, - 2) . " " . $kga['lang']['export_extension']['duration_unit'];
     }
 
     /**
--- a/extensions/ki_export/base_export_pdf.php
+++ b/extensions/ki_export/base_export_pdf.php
@@ -109,7 +109,8 @@ class BasePDF extends TCPDF
         $s = $duration % 60;
         $m = (($duration - $s) / 60) % 60;
         $h = floor($duration / 3600);
-        return $h . ":" . substr('0' . $m, - 2) . " " . $kga['lang']['export_extension']['duration_unit'];
+        $timeLength =  $h . ":" . substr('0' . $m, - 2);
+        return $this->time_unit($timeLength);
     }
 
     /**
--- a/extensions/ki_export/base_export_pdf.php
+++ b/extensions/ki_export/base_export_pdf.php
@@ -96,6 +96,20 @@ class BasePDF extends TCPDF
 
         return sprintf('% 2d:%02d', $hours, $minutes);
     }
+    
+    /**
+     * calculate time in hh:mm from duration lenght in second
+     * rounded to minutes
+     * @param int $duration, duration lenght in second
+     * @return string time in format hh:mm
+     */
+    public function timeLength($duration)
+    {
+        $s=$duration % 60;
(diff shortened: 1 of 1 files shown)


claudegel:
This was part of my export to pdf2 PR that have not been merged for some reason.

simonschaufi:
Please format your code according to PSR standards.

claudegel:
I've done some formating. Don't know what to do more. My code look like the other lines. My formating also removed few trailing spaces, sorry

simonschaufi:
do you use phpstorm? there is a setting for PSR code formating

claudegel:
I'm using zend studio 13 but the PSR formating seems to be broken.
I should try phpstorm. Didnyou told me that you had free licenses for phpstorm?

Claude Gelinas agr.
Phyto Ressources Inc.

claudegel:
Zend Studio use PSR-2 instead of PSR formatting. Is it the same ?

simonschaufi:
Yes, PSR-2 is the way to go.

claudegel:
Ok my code is formatted against PSR-2

extensions/ki_export/base_export_pdf.php
```diff
@@ -98,6 +98,22 @@ public function SumStdTime($time1, $timesum)
...
+     * @param int $duration
+     *            , duration lenght in second
+     * @return string time in format hh:mm
+     */
+    public function timeLength($duration)
+    {
+        $s = $duration % 60;
+        $m = (($duration - $s) / 60) % 60;
+        $h = floor($duration / 3600);
+        return $h . ":" . substr("0" . $m, - 2) . " " . $kga['lang']['export_extension']['duration_unit'];
```
simonschaufi: please use single quotes here.

extensions/ki_export/base_export_pdf.php
```diff
@@ -98,6 +98,22 @@ public function SumStdTime($time1, $timesum)
     }
 
     /**
+     * Calculate time in hh:mm from duration lenght in second
+     * rounded to minutes.
+     *
+     * @param int $duration
+     *            , duration lenght in second
```
simonschaufi: that line is a bit messed up. should belong to line 104

extensions/ki_export/base_export_pdf.php
```diff
@@ -98,6 +98,21 @@ public function SumStdTime($time1, $timesum)
...
+     *
+     * @param int $duration , duration lenght in second
+     * @return string time in format hh:mm
+     */
+    public function timeLength($duration)
+    {
+        $s = $duration % 60;
+        $m = (($duration - $s) / 60) % 60;
+        $h = floor($duration / 3600);
+        return $h . ":" . substr('0' . $m, - 2) . " " . $kga['lang']['export_extension']['duration_unit'];
```
kevinpapst: Did you check that it works? No log entry? $kga should not be accessible here?!?

I would prefer:

```
$timeLength =  $h . ":" . substr('0' . $m, - 2);
return $this->time_unit($timeLength);
```

extensions/ki_export/base_export_pdf.php
```diff
@@ -200,7 +215,7 @@ public function printSummary($header, $data)
                 }
             } else {
                 if (isset($this->columns['time'])) {
-                    $this->Cell($w[1], 6, gmdate('H:i', $row['std_time']), 'LR', 0, 'R', $fill);
+                    $this->Cell($w[1], 6, $this->timeLength($row['std_time']), 'LR', 0, 'R', $fill);
```
kevinpapst, line 219: The cell did not include the duration unit translation before. I am just bringing this up, because from my POV the unit is not really necessary here. But I am open for discussions ;-)
Why did you choose to include it?

claudegel:
Because the duration unit is included for the detailed list and the cost unit ($) is displayed in both summary and detail, I think it is logic to add the duration unit in the summary. It is also included in the summary if we chose decimal time.

also I tested and it is better to go with:

``` php
$timeLength =  $h . ":" . substr('0' . $m, - 2);
return $this->time_unit($timeLength);
```

I'll make the change in the PR for that

extensions/ki_export/base_export_pdf.php
```diff
@@ -98,6 +98,22 @@ public function SumStdTime($time1, $timesum)
     }
 
     /**
+     * Calculate time in hh:mm from duration lenght in second
+     * rounded to minutes.
+     *
+     * @param int $duration , duration lenght in second
```
simonschaufi: remove the comma and just make the coment: "lenght in seconds"

kevinpapst:
Thanks!

PR #784, opened 2016-08-18
Auth config

### LINKED ISSUE #760: Subclass of ldapadvanced fails to authenticate locally
#### Expected behaviour
Login with local user "admin"
#### Actual behaviour
HTTP Error 500 and no login with local user
Create subclass of ldapadvanced and use it as authenticator.
Edits of ldapadvanced were revoked to original version
Edits of ldapadvanced were kept for our use case
Both times we get HTTP Error 500 while using the subclass as authenticator. Everything works fine by editing ldapadvanced.php for our needs and authenticate locally.
(refered to: https://github.com/kimai/kimai/issues/758)
#### Kimai version
1.0.1.1388
none. Can't login to see tab.
#### Relevant log entries from Apache Error Log
PHP Fatal error:  Call to a member function authenticate() on a non-object in /var/www/kimai/libraries/Kimai/Auth/MyAuthenticator.php on line 204,

kevinpapst:
FIXES #760

- allow to configure each authenticator class with a config file

Reason for this pull request:
- Kimai updates would overwrite user settings made in e.g. https://github.com/kimai/kimai/blob/develop/libraries/Kimai/Auth/Ldapadvanced.php


kevinpapst:
ups, wrong target branch

PR #788, opened 2016-08-19
[FEATURE] add Indesign XML Export

### LINKED ISSUE #783: ENHANCEMENT Indesign CS6 XML Export (function)
Dear Kimaiers

I'm mainly a cross field designer-maker based on video.
But I'm using Kimai for years now and lately I was fed up with my patchy XLS to PDF workflow. It really looks bad in my Indesign Document.

I made a function for Kimai which produces an XML file that Indesign can import natively. Based on that you can define your formatting with fonts and cells and stuff and hook them to the tags used in the XML (tagnames freely chosen).
So you have a personalized Indd Template and any new XML export will be formatted on-the-fly (in a beautiful meaning).

I really never used GIT before, and have no time now to learn it. So please excuse me if I violently mistreat your guidelines ;-)
I thought since it's a whole function and you know so much more than me on how to implement it, it's fair to just pass it over. Thank you.

Ähem another thing is... My version of Kimai was not exactly recent, but I hope I'm not too far off the DB. I built it on top the XLS array. There's a reference in the comments.
I cannot update my Kimai right now, as I build another "plugin" for a click-click iphone insert straight into the DB. So I'm quite stuck for the present.

See the comments for details, version, TODO etc.
More specific comments are also inline.

Good luck with your project.

Best from se Schwiitzerlend
Manu
[private_func_inddXML.php.zip](https://github.com/kimai/kimai/files/423733/private_func_inddXML.php.zip)

simonschaufi:
FIXES #783


labor4:
yibee

kevinpapst:
How can this be tested?

simonschaufi:
well, since i also don't have indesign, i just checked if there is a "meaningful" xml rendered

labor4:
> How can this be tested?

Procedure:
- In Indesign: Import XML
- The structure appears as a tree in the "Tags" tool
- drag the "ProtocolHolder" container into a sheet
- the table is visible without formatting

then:
- produce cell formats, header formats etc, include paragraph formats, make text formats for strings like "ProtocCustTitle.CUSTSTRING"
- link the tags to the formats (in the "Tags" tool)
- save Indd doc as master, delete the table clean in the "Tags" tool, and reimport to check the format links (autoformatting)
- reuse the Indd Master or create an Indd template off it.

Formats can also be imported from other Indd documents.
The full structure is visible if you export multiple Customers with multiple Projects including worktime and expenses. The data will also be sorted this way.

m

btw:
[example.pdf](https://github.com/kimai/kimai/files/423808/example.pdf)

kevinpapst:
example.pdf looks good. 
But to be honest: I don't think that this should be part of Kimai core.
Indesign is not the most wide spread software, compared to PDF, Excel or Open Office.
We should wait, until we have a better Plugin system and then release this as official plugin.

labor4:
yes. technically this is true. 
but neither of these target creatives. 
(i mean supposed i need to present my protocol to customers)
i see the effort around openoffice, but i myself would prefer not to try to design in such a program. its based on Word (different purpose).
what i did not research: indesign uses the xml standard as a machine interchangeable container for large data inputs. so i wouldnt be surprised if another app can work on it. 
but this i cannot reference just now. 

m

labor4:
think of my example as the highest level of concept, a designer can still accept to work with freely.

labor4:
> what i did not research: indesign uses the xml standard as a machine interchangeable container for large data inputs. so i wouldnt be surprised if another app can work on it. 
> but this i cannot reference just now.

No.

i downloaded OpenOffice, LibreOffice, Scribus, and also checked my M$ Office:
It seems none of them can open this XML. M$ said it would be ok with an additional XSLT stylesheet, but giving an empty one errored (quick test).
BTW I tried an export of M$ Word XML, which cannot be read by LibreOffice or OpenOffice.

It seems my XML script is Indesign-only for now.
Probably aid:table="cell": ........ "aid" rather stands for AdobeInDesign, than an actual aid... :-(

m

simonschaufi:
closing this as we won't do huge refactorings any more. please use kimai 2 where own exports formats are easily possible

PR #789, opened 2016-08-20
Authenticator: Exception if LDAP extension is not loaded

kevinpapst:
- throw exception if LDAP extension is not loaded
- phpdoc improvements

Reason for this pull request:
- LDAP classes should behave the same way (see Ldapadvanced)

### DIFF
--- a/libraries/Kimai/Auth/Ldap.php
+++ b/libraries/Kimai/Auth/Ldap.php
@@ -62,14 +62,12 @@ class Kimai_Auth_Ldap extends Kimai_Auth_Abstract
     private $kimaiAuth = null;
 
     /**
-     * Kimai_Auth_Ldap constructor.
-     * @param null $database
-     * @param null $kga
+     * {@inherit}
      */
     public function __construct($database = null, $kga = null)
     {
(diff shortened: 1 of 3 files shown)


libraries/Kimai/Auth/Ldap.php
```diff
@@ -25,43 +25,69 @@
...
 
+    /**
+     * Kimai_Auth_Ldap constructor.
+     * @param null $database
+     * @param null $kga
+     */
     public function __construct($database = null, $kga = null)
     {
+        if (!function_exists('ldap_bind')) {
+            throw new InvalidArgumentException('LDAP-Extension is not installed');
```
simonschaufi: I think InvalidArgumentException is not the right type of exception to be thrown. There is no argument which is invalid. I'm voting for RuntimeException

kevinpapst: Copy & Paste from https://github.com/kimai/kimai/blob/master/libraries/Kimai/Auth/Ldapadvanced.php#L184 
but you are right, I will update both places to throw the same connection.

Issue #738, opened 2016-07-12
Cant edit configuration - configuration_edit: Column 'value' cannot be null


tiMon1987:
Hey,
im on it now for the second day. I wanted to test the latest Kimai Version on our webserver. Installation works fine. But I am not able to change the configuration. Seems to be a 'not null' problem, because it tries to set NULL for all unchecked checkboxes.
The "NOT NULL" bit for "value" seems to be correct when i look at the install php:

```
$query =
"CREATE TABLE `${p}configuration` (
  `option` varchar(255) NOT NULL,
  `value` varchar(255) NOT NULL,
  PRIMARY KEY  (`option`)
);";
exec_query($query);
```
If go to the database and allow NULL for the value-column i can change the config in kimai. The default 0s for the unchecked checkboxes are then overritten with NULL.

#### Expected behaviour
Edit the the Configuration
#### Actual behaviour
Not able to edit - getting "configuration_edit: Column 'value' cannot be null (#1048)"
#### Steps to reproduce the behaviour
Clean install of newest Version and change any configuration value (it need not to be the checkbox)
#### Kimai version
latest stable 1.0.1 and same behavior with latest masterbranch
#### PHP/mySQL Version:
PHP Version 5.6.20-0 / mySQL Version 5.7.11
#### Relevant log entries from Kimais Debug tab
> [11.07.2016 13:48:57] -- begin install ----------------------------------
> [11.07.2016 13:49:00] -- installation finished without error ------------
> [11.07.2016 13:49:04] login: admin as user
> [11.07.2016 13:49:11] KSPI axAction (admin): securityWarning
> [11.07.2016 13:49:14] KSPI axAction (admin): refreshSubtab
> [11.07.2016 13:49:14] KSPI axAction (admin): refreshSubtab
> [11.07.2016 13:49:14] KSPI axAction (admin): refreshSubtab
> [11.07.2016 13:49:43] KSPI axAction (admin): sendEditAdvanced
> [11.07.2016 13:49:43] configuration_edit: Column 'value' cannot be null (#1048)

I get similar problems when i try to add a user and set a password:
> [12.07.2016 10:53:29] KSPI axAction (admin): createUser
> [12.07.2016 10:53:30] KSPI axAction (admin): editUser
> [12.07.2016 10:53:30] KSPI axAction (admin): refreshSubtab
> [12.07.2016 10:53:34] KSPI axAction (admin): sendEditUser
> [12.07.2016 10:53:34] user_edit failed: 
[...]

PR #792, opened 2016-08-20
[BUGFIX] Column 'value' cannot be null

simonschaufi:
In getRequestBool the value was checked like this:

``` php
$_REQUEST[$_REQUEST[$name]]
```

but it should be checked like this:

``` php
$_REQUEST[$name]
```

FIXES: #738


kevinpapst:
Good catch!!!! Hope that fixes the Problem.

Issue #721, opened 2016-06-23
Make time format configurable in timesheet table


simonschaufi:
#### Expected behaviour
As a user I want to configure the format of the time in the timesheet table (am|pm). This should be user independent (in user settings, not in admin panel)

#### Actual behaviour
currently hardcoded: https://github.com/kimai/kimai/blob/master/extensions/ki_timesheets/templates/scripts/timeSheet.php#L128

PR #793, opened 2016-08-20
[FEATURE] Make time format configurable in timesheet table

simonschaufi:
Resolves: #721 
- Add a new field in admin panel for global table time setting
- Add field in user preferences to overwrite global table time setting


kevinpapst:
Also conflicts with #688

Issue #636, opened 2016-03-24
use utf8 as default character set


simonschaufi:
currently the database is created as latin1.
the important line is this one:
https://github.com/kimai/kimai/blob/master/includes/vars.php#L35
new installations should be created with utf8 while current installations should be untouched. there should be a config setting written to autoconf to set the charset.

PR #794, opened 2016-08-21
[FEATURE] use utf8 as default character set

simonschaufi:
Fixes: #636

### DIFF
--- a/libraries/Kimai/Database/Mysql.php
+++ b/libraries/Kimai/Database/Mysql.php
@@ -49,12 +49,16 @@ class Kimai_Database_Mysql
     {
         $this->kga = &$kga;
         if ($autoConnect) {
+            $useUtf8 = false;
+            if ($kga['utf8'] || $kga['server_charset'] === 'utf8') {
+                $useUtf8 = true;
+            }
             $this->connect(
                 $kga['server_hostname'],
                 $kga['server_database'],
                 $kga['server_username'],
                 $kga['server_password'],
-                $kga['utf8']
+                $useUtf8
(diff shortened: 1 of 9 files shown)


simonschaufi:
Don't merge yet.

kevinpapst:
That one also conflicts massively with KGA object

libraries/Kimai/Database/Mysql.php
```diff
@@ -49,12 +49,16 @@ public function __construct(&$kga, $autoConnect = true)
     {
         $this->kga = &$kga;
         if ($autoConnect) {
+            $useUtf8 = false;
+            if ($kga['utf8'] || $kga['server_charset'] === 'utf8') {
```
kevinpapst: I understand that we shouldn't simply remove the kga['utf8'] setting for backward compatibility.
But in a couple of month, we will have forgotten, why we have to settings for the same feature.

simonschaufi: The problem are the existing users who already use utf8 and update kimai ($kga['server_charset'] will NOT be utf8 in that case because of this line: https://github.com/simonschaufi/kimai/blob/6e0bec8d564e954b497311cd9c88947782665110/includes/vars.php#L48)

kevinpapst: Yes true, I was just thinking out loud. Probably we can find a way during the update to check that and set the variables accordingly?!?

simonschaufi: yes, right.

PR #795, opened 2016-08-21
[FEATURE] Central place for password encoding

simonschaufi:
Relates: #378

### DIFF
--- a/core/processor.php
+++ b/core/processor.php
@@ -100,7 +100,7 @@ switch ($axAction) {
 
         // If the password field is empty don't overwrite the old password.
         if (trim($_REQUEST['password']) != "") {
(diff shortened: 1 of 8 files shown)


kevinpapst:
Hey, good one!
But I was still thinking about it ... actually I would prefer, if we put that function into the Authenticator - but that can be done in a later step. Now its much easier to refactor that code.
Thanks!

simonschaufi:
I also thought about putting it there but then decided to put it in functions.php simply because there is no auth object easily available and static methods are not good for unit tests (as i heard)

kevinpapst:
Will be, once we merge KGA to Object :)

2016-08-23 1:28 GMT+02:00 Simon Schaufelberger notifications@github.com:

Issue #796, opened 2016-08-22
Projects cannot be created with new Tasks assigned


Ruckhaber:
I wanted to create a new project in the admin-Panel. I added some new activities. After pressing the OK button nothing happens. The project is not created. Only the CANCEL button works .

#### Reproducable on demo.kimai.org?
yes

#### Kimai version
dev 1.0.1.1388, github develop

simonschaufi:
please don't modify our issue template! Add your own changes BELOW our lines!

simonschaufi:
```
Request URL:http://demo.kimai.org/core/processor.php
Request Method:POST
Status Code:500 Internal Server Error
```
this would be a helpful information!

PR #798, opened 2016-08-22
[BUGFIX] New projects cannot be added with tasks assigned

simonschaufi:
Return empty array instead of false

Fixes: #796


kevinpapst:
Could you please stop pushing & merging so quickly?
Give us some time to review... even if "us" means I ;-)

simonschaufi:
yes, now that i released the new version, the "normal" review process takes place ;)

PR #799, opened 2016-08-22
[BUGFIX] Remove all assignments with activities when none is given

simonschaufi:


kevinpapst:
Hey, why do you merge so fast? Any hurry?

simonschaufi:
yes, i just wanted to kick out a new release and that bug came from me :D

kevinpapst:
Yeah, well ... I am not a friend of that. There is really no need to hurry
or to push last minute changes into a release.

PR #800, opened 2016-08-22
[BUGFIX] Avoid warning in foreach

simonschaufi:


core/floaters.php
```diff
@@ -248,9 +248,11 @@
                 $view->assign('selectedProjectIds', $selectedProjectIds);
 
                 $selectedProjects = $database->activity_get_projects($id);
-                foreach ($selectedProjects as &$selectedProject) {
-                    // edit by reference!
-                    $selectedProject['fixedRate'] = $database->get_fixed_rate($selectedProject['projectID'], $id);
+                if (is_array($selectedProjects)) {
+                    foreach ($selectedProjects as &$selectedProject) {
```
kevinpapst, line 252: Please remove the copy&pasted & as this is not required with PHP 5, so the code gets even better :)

simonschaufi: could you make a new PR please?

kevinpapst: I searched the codebase, there are multiple places with this kind of reference manipulation.
We need to keep an eye on it, especially for PHP 7. We had some problems with this kind of code in a large business project.

PR #802, opened 2016-08-23
Support cyrillic character

### LINKED ISSUE #801: Something wrong with cyrillic symbols in exported pdf
There is something wrong with cyrillic in exported pdf

screen shot 2016-08-23 at 20 13 50
#### Steps to reproduce
1. Use any cyrillic symbols in nicknames/projects/customers/anywhere
2. Export PDF via "Export" tab
#### Reproducable on demo.kimai.org? [yes|no]

Yes, easily reproducable
#### Kimai version [for example: dev 1.1.0.1389|github master|github develop]

github develop

kevinpapst:
FIXES #801

- use the font "freesans" to display cyrillic character (instead of helvetica)

### DIFF
--- a/extensions/ki_export/export_pdf.php
+++ b/extensions/ki_export/export_pdf.php
@@ -40,17 +40,17 @@ class MYPDF extends BasePDF
         $this->SetY(-15);
 
         // customer data
-        //$this->SetFont('helvetica', '', 8); // Set font
+        //$this->SetFont('freesans', '', 8); // Set font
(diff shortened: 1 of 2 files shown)


kevinpapst:
@simonschaufi This does not fix the Ruble problem, but it fixes a lot of others UTF-8 problems.
Shall we merge it, so the export PDF is working better as before? Even if the feature is not 100% - it is far better than what we have now.

simonschaufi:
better a 5% broken system than a 90% :D

http://stackoverflow.com/questions/20665622/russian-ruble-symbol-html-code ;)

PR #704, opened 2016-05-13
export in dec_time or std_time in export_pdf2

claudegel:
Changes proposed in this pull request:
- Add the possibility to export in standard time or decimal time in the pdf2 export extention

Reason for this pull request:
- This PR replace two old PR to rebase them to master, #544 and #543

PR #810, opened 2016-08-28
export in dec_time or std_time in export_pdf2

claudegel:
FIXES #

- Add the possibility to export in standard time or decimal time in the pdf2 export extension

Reason for this pull request:
- This PR replace old PR #704 to rebase it against develop branch. this time it look ok

### DIFF
--- a/extensions/ki_export/export_pdf2.php
+++ b/extensions/ki_export/export_pdf2.php
@@ -65,14 +65,14 @@ class MYPDF extends BasePDF {
   public function Footer() {
     global $kga, $customerData, $projectData;
 
-    // Position at 1.5 cm from bottom 
+    // Position at 1.5 cm from bottom
     $this->SetY(-15);
 
     // customer data
     /*$this->SetFont('freesans', '', 8); // Set font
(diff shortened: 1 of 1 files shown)


extensions/ki_export/export_pdf2.php
```diff
@@ -102,7 +102,16 @@ public function printRows($data, $widths) {
       else {
         $this->printTimeRow($widths, $row);
         $this->moneySum += $row['wage'];
-        $this->timeSum += $row['decimalDuration'] == -1 ? 0 : $row['decimalDuration'];
+        if ($_REQUEST['time_type'] == "dec_time") {
```
kevinpapst, line 105: It looks as if there are race-conditions where timeSum will not be generated. 
I would prefer creating a default timeSum (leaving the previous line) and then overwriting its value in case `$_REQUEST['time_type'] != 'dec_time' && isset($this->columns['time'])`

claudegel: on line 105 if $_REQUEST['time_type'] != 'dec_time then it goes to line 109 } else {
and there, if isset($this->columns['time'] TimeSum is updated
I don't understand your race-condition

PR #819, opened 2016-09-02
[BUGFIX] only call updater when database changes

simonschaufi:
- call the updater only when database changes

Reason for this pull request:
- updater is also called when kimai version changes without database changes. This will result in an endless loop.

### DIFF
--- a/includes/func.php
+++ b/includes/func.php
@@ -235,7 +235,6 @@ function checkDBversion($path)
     // check for versions before 0.7.13r96
     $installedVersion = $database->get_DBversion();
     $checkVersion = $installedVersion[0];
-    $checkVersion = "$checkVersion";
 
     if ($checkVersion == "0.5.1" && count($database->get_users()) == 0) {
         // fresh install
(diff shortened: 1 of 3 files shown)


includes/func.php
```diff
@@ -235,15 +235,15 @@ function checkDBversion($path)
...
 
     if ($checkVersion == "0.5.1" && count($database->get_users()) == 0) {
         // fresh install
         header("Location: $path/installer/");
         exit;
     }
 
-    if ($checkVersion != $config->getVersion() || (int)$installedVersion[1] < $config->getRevision()) {
+    // only call updater when database changes no matter the kimai version
+    if ((int)$installedVersion[1] < $config->getRevision()) {
```
kevinpapst, line 246: Imagine the updater would delete a file, but the database didn't change. 
Are we still going to increase the revision?

simonschaufi: Well, If we just increase the version number, we get a redirect loop to the updater. The problem is this line where the kimai version only gets set when the revision changes.
https://github.com/kimai/kimai/blob/develop/updater/updater.php#L1139
When we think a version depending migration also makes sense, then we need to revert that and fix the updater.

Issue #812, opened 2016-08-29
Identical project names


alsopeter:
Let me start by saying, nice job!
Finally time tracking software that does what you need, and nothing more.

I have been using it for a week now, and there is one issue from witch I don’t know if I am doing it wrong, or it is a nice to have. For some customers I do projects, but also some general business that can only be named as general.  As project I connect general as project name to this customer and also call the activity general. Because I have multiple customers, I do general work for, the project column contains multiple “general” projects. The activities for these project are also named general.  Kimai is capable of connecting the general activity to the right general project and connect it to the customer. 

Problem occurs when for a new customer, I insert a general project with general activity. I create the new customer, create general project and connect it to this customer. In the downright activities box I push add activity and the box pops up. I insert “general” as activity name. On the projects tab, I can connect this activity to the new added general project via the pulldown, but this is where the problem occurs. There are multiple general projects and it is not clear to witch customer the general projects are connected.

So now I just select one, and via filtering I check if it is the correct one. Keep changing, until I have the correct project.

Am I doing something wrong, or is it a nice to have for the add activity popup to see to witch customer the projects in the pulldown belong to?

simonschaufi:
Indeed it would be nice to have the customer name mentioned as well. I also noticed that already. Thank you for your inspiration!
PS: You can also edit the project and connect the activity there.

alsopeter:
Done. found it in add_edit_activity.php
Why didn't I think of that earlier?
Perhaps because it was monday morning :-)

Thank!

simonschaufi:
Wanna do a Pull Request? That would be nice!

alsopeter:
Never did a pull Request before on github.
First have to find out what it exactly is :-)

simonschaufi:
Click on fork, then you have your own copy of kimai. create a new branch, make your changes, push your branch and then make a pull request via github.

kevinpapst:
added in develop

PR #825, opened 2016-09-05
[FEATURE] Add customer name as additional info

simonschaufi:
Fixes: #812


kevinpapst (approved):
I really hate our database layer ... customerName vs customer_name ... I wish we had real entities^^

PR #829, opened 2016-09-07
Adding optional 'description' parameter on startRecord and stopRecord API

christmasso:
- adding an optional parameters on startRecord and stopRecord API. The parameter contains the description of the related activity.

Reason for this pull request:
- a mobile version of kimai with the possibility to add the description can be useful (see also the feature request 'Add description field' on kimai-mobile).
- in the following days I will release a new native kimai Android app with GPL license that will use the modified API. Now the Android app is in beta-testing.

### DIFF
--- a/libraries/Kimai/Database/Mysql.php
+++ b/libraries/Kimai/Database/Mysql.php
@@ -3708,11 +3708,10 @@ class Kimai_Database_Mysql
      * Performed when the stop buzzer is hit.
      *
      * @param integer $id id of the entry to stop
-     * @param string $description associated to the entry to stop
      * @author th, sl
      * @return boolean
      */
-    public function stopRecorder($id, $description = null)
+    public function stopRecorder($id)
     {
         $table = $this->getTimeSheetTable();
 
@@ -3731,9 +3730,6 @@ class Kimai_Database_Mysql
         $values['start'] = $rounded['start'];
         $values['end'] = $rounded['end'];
         $values['duration'] = $values['end'] - $values['start'];
-        if ($description != null) {
-            $values['description'] = MySQL::SQLValue($description);
-        }
 
         $query = MySQL::BuildSQLUpdate($table, $values, $filter);
 
@@ -3746,11 +3742,10 @@ class Kimai_Database_Mysql
      * @param integer $projectID ID of project to record
      * @param $activityID
      * @param $user
-     * @param string $description associated to the entry to stop
      * @return int id of the new entry or false on failure
      * @author th, sl
      */
-    public function startRecorder($projectID, $activityID, $user, $description = null)
+    public function startRecorder($projectID, $activityID, $user)
     {
         $projectID = MySQL::SQLValue($projectID, MySQL::SQLVALUE_NUMBER);
         $activityID = MySQL::SQLValue($activityID, MySQL::SQLVALUE_NUMBER);
@@ -3761,10 +3756,6 @@ class Kimai_Database_Mysql
         $values['start'] = time();
         $values['userID'] = $user;
         $values['statusID'] = $this->kga->getDefaultStatus();
-        if ($description != null) {
-            $values['description'] = MySQL::SQLValue($description);
-        }
-
         $rate = $this->get_best_fitting_rate($user, $projectID, $activityID);
         if ($rate) {
             $values['rate'] = $rate;
--- a/libraries/Kimai/Database/Mysql.php
+++ b/libraries/Kimai/Database/Mysql.php
@@ -3708,10 +3708,11 @@ class Kimai_Database_Mysql
      * Performed when the stop buzzer is hit.
      *
      * @param integer $id id of the entry to stop
+     * @param string $description associated to the entry to stop
      * @author th, sl
      * @return boolean
      */
-    public function stopRecorder($id)
+    public function stopRecorder($id, $description = null)
     {
         $table = $this->getTimeSheetTable();
 
@@ -3730,6 +3731,9 @@ class Kimai_Database_Mysql
         $values['start'] = $rounded['start'];
         $values['end'] = $rounded['end'];
         $values['duration'] = $values['end'] - $values['start'];
+        if ($description != null) {
+            $values['description'] = MySQL::SQLValue($description);
+        }
 
         $query = MySQL::BuildSQLUpdate($table, $values, $filter);
 
@@ -3742,10 +3746,11 @@ class Kimai_Database_Mysql
      * @param integer $projectID ID of project to record
      * @param $activityID
      * @param $user
+     * @param string $description associated to the entry to stop
      * @return int id of the new entry or false on failure
      * @author th, sl
      */
-    public function startRecorder($projectID, $activityID, $user)
+    public function startRecorder($projectID, $activityID, $user, $description = null)
     {
         $projectID = MySQL::SQLValue($projectID, MySQL::SQLVALUE_NUMBER);
         $activityID = MySQL::SQLValue($activityID, MySQL::SQLVALUE_NUMBER);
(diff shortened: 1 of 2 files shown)


kevinpapst:
Hey, I really like the idea for a native GPL app. Fantastic News!!!
If you need beta tester, I am here to help :-)

Regarding your PR: the changes might be logical from a UI/UX perspective, but from an API perspective they are not the way I would design them.
Please try to implement it with separate calls. One to start/stop and one for updating an entry.
For example: The next person wants to set the location in startRecorder and doesn't need  the description, what now?

kevinpapst:
Ups, github mobile view is complicated ;-) Accidentaly closed: reopen.

christmasso:
Perfect, I'm happy to add you as beta tester! :-)

To clarify the problem related to the API design: do you think it is more appropriate to have various calls for the single parameters (like setDescription, setLocation, etc...) or have an optional $data object passed to the start/stop calls managing all the allowed parameters?

simonschaufi:
I would also like to have access to the native app!
I think 2 calls is not the best idea. I would vote for the second option with the optional data array.

kevinpapst:
I would say: leave the stop function as It is. The optional data array
could be placed in the start function and should support as many fields as
possible.

From an API perspective there should be a function to update an entry,
which could be called internally by the start function.

What do you need from us to get a beta apk?

christmasso:
Sounds good, I'll proceed with your advices.

I'll need your android/gmail account to add you as a test members, so that you can see the beta version of the app from the google play store.

christmasso:
I have added an update entry function API, that will also be called by the start recording if optional data are passed.
It uses the already written timeEntry_edit (in Mysql.php) to update data on DB, since the logic needed is the same.

@kevinpapst I'm writing to your personal mail for details about the beta testing. Please share it with Simon, I didn't immediately find his email address :-)

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
         }
         */
 
-        $result = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
-        if ($result) {
+        $entryId = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
```
simonschaufi, line 269: instead of creating  new record here and update it a few lines later, I would suggest passing the data here already and process it right in the startRecorder method. @kevinpapst what do you think?

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
...
         }
         return $this->getErrorResult();
     }
 
+    /**
+     * Updates the currently running recording. If $entryId is empty the
+     * current activity will be updated.
+     *
+     * @param string $apiKey
+     * @param integer $entryId
```
simonschaufi, line 287: please use `int`

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
...
+
+            // no "last" activity existing
+            if (count($result) == 0) {
+                return $this->getErrorResult('No active recording.');
+            }
+
+            // get the data of the first active recording
+            $result = $this->getBackend()->timeSheet_get_data($result[0]);
+
+            if (array_key_exists('timeEntryID', $result)) {
```
simonschaufi, line 311: `if (isset($result['timeEntryID']))`

libraries/Kimai/Database/Mysql.php
```diff
@@ -3756,7 +3756,6 @@ public function startRecorder($projectID, $activityID, $user)
         $values['start'] = time();
         $values['userID'] = $user;
         $values['statusID'] = $this->kga->getDefaultStatus();
-
```
simonschaufi, line 3759: please leave it like it was.

simonschaufi (changes requested):

libraries/Kimai/Remote/Api.php
```diff
@@ -231,9 +231,10 @@ protected function getSuccessResult(array $items, $total = 0)
      * @param string $apiKey
      * @param integer $projectId
      * @param integer $activityId
+     * @param array $data
      * @return array
      */
-    public function startRecord($apiKey, $projectId, $activityId)
+    public function startRecord($apiKey, $projectId, $activityId, $data = null)
```
kevinpapst, line 237: an array should be initialized with `$data = array()` instead of `$data = null` ...

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
         }
         */
 
-        $result = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
-        if ($result) {
+        $entryId = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
+        if ($entryId) {
+            // if there are optional data we update the record entry
+            if (!is_null($data)) {
```
kevinpapst, line 272: ... and here a check for `if (!empty($data))`

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
...
+     * @return boolean
+     */
+    public function updateEntry($apiKey, $entryId, $record)
+    {
+        if (!$this->init($apiKey, 'updateEntry')) {
+            return $this->getAuthErrorResult();
+        }
+
+        // check for empty params
+        if ($entryId == '') {
```
kevinpapst, line 298: Return early:
```
if (empty($entryId)) {
    return $this->getErrorResult('Invalid entry ID given');
}
```

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
...
+    /**
+     * Updates the currently running recording. If $entryId is empty the
+     * current activity will be updated.
+     *
+     * @param string $apiKey
+     * @param integer $entryId
+     * @param array $record
+     * @return boolean
+     */
+    public function updateEntry($apiKey, $entryId, $record)
```
kevinpapst, line 291: Please rename `$record` to `$data` as it is misleading - I first thought this is a database record

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
...
+        }
+        
+        // valid $record?
+        if (empty($record)) {
+            return $this->getErrorResult('Invalid record');
+        }
+
+        // optional values
+        if (isset($record['location'])) {
+            $data['location'] = $record['location'];
```
kevinpapst, line 323: `$data` should then be renamed to `$updateValues` and please initialize it before using: 
`$updateValues = array();`

libraries/Kimai/Remote/Api.php
```diff
@@ -265,15 +266,103 @@ public function startRecord($apiKey, $projectId, $activityId)
         }
         */
 
-        $result = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
-        if ($result) {
+        $entryId = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
+        if ($entryId) {
+            // if there are optional data we update the record entry
+            if (!is_null($data)) {
+                $this->updateEntry($apiKey, $entryId, $data);
```
kevinpapst, line 273: You need to check the result from that call. If you passed in data and the update of the new entry failed, then the API consumer should be notified about it.

kevinpapst (changes requested):

kevinpapst:
Thanks for your contribution @christmasso 👍 - I added some comments, please have a look.

kevinpapst:
Oha, just saw how old this PR is... we must have missed it.
@christmasso do you still want to work on it or should it be closed?

simonschaufi:
@christmasso any update?

simonschaufi:
No response and won't fix.

Issue #787, opened 2016-08-19
Add record via +: wrong hourly rate


bgeiser:
#### Expected behaviour
The hourly rate assigned to a project is shown in the budget tab and correctly recorded.

#### Actual behaviour
For (internal) projects with 0 hourly rate the rate of a previously selected project is shown in the budget tab and recorded.

#### Steps to reproduce
Add record to time sheet by clicking on +. Select a project with an hourly rate >0. Then select project with hourly rate ==0. Click on budget tab. The previous rate >0 is still shown and will be recorded

#### Reproducable on demo.kimai.org? [yes|no]
yes

#### Kimai version [for example: dev 1.0.1.1388|github master|github develop]
v1.0.1.1388

kevinpapst:
fixed in develop

PR #838, opened 2016-09-20
[BUGFIX] FixedRate is not considered while adding a new entry manually

simonschaufi:
Fixes: #787

### DIFF
--- a/libraries/Kimai/Database/Mysql.php
+++ b/libraries/Kimai/Database/Mysql.php
@@ -2285,27 +2285,28 @@ class Kimai_Database_Mysql
     {
         $data = $this->clean_data($data);
 
-        $values['location']     = MySQL::SQLValue($data['location']);
-        $values['comment']      = MySQL::SQLValue($data['comment']);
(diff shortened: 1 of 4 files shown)


libraries/Kimai/Database/Mysql.php
```diff
@@ -4521,10 +4522,10 @@ public function get_best_fitting_rate($userID, $projectID, $activityID)
             return false;
         }
 
-        if ($this->conn->RowCount() == 0) {
+        if ($this->conn->RowCount() == -1) {
```
kevinpapst, line 4525: <= 0 should be the best. the == 0 check worked in the past. did you activate strict mode or why did that change?

kevinpapst (commented):
I don't know what you changed, all these style changes make it impossible for me to review.

kevinpapst (approved):

libraries/Kimai/Database/Mysql.php
simonschaufi: no, this has nothing to do with strict mode. It's just the behavior of the MySql Class

PR #863, opened 2016-11-11
Dutch language improvements

Hillefuego:
Made improvements to the Dutch language file.
- Fixed typo's
- Copied keywords from en.php file and translated the descriptions.


simonschaufi (approved):

language/nl.php
```diff
@@ -22,7 +22,7 @@
  * by Paul Brand and Vasilis van Gemert
  */
 return array(
-    "countryCode" => "dutch",
+    "countryCode" => "Dutch",
```
simonschaufi: the countryCode is definitely wrong (was also wrong before). This should be nl-NL

PR #866, opened 2016-11-13
[FEATURE] Introduce Kimai API

simonschaufi:
With this class Kimai data can be fetched with a clean database API. More methods will be added over time.

How to use:
```php
require_once 'path/to/kimai/libraries/Kimai/Api.php';

$api = new \Kimai_Api([
    'key' => 'api key of user',
    'path' => '/path/to/kimai'
]);

$customers = $api->getCustomers();

echo '<pre>' . print_r($customers, true) . '</pre>';
```

### DIFF
--- a/libraries/Kimai/Api.php
+++ b/libraries/Kimai/Api.php
@@ -1,0 +1,155 @@
+<?php
+
+/**
+ * Class Kimai_Api
+ *
+ * Example:
+ *
+ * require_once 'path/to/kimai/libraries/Kimai/Api.php';
+ * $api = new \Kimai_Api([
+ *     'key' => 'api key of user',
+ *     'path' => '/path/to/kimai'
+ * ]);
+ * $customers = $api->getCustomers();
+ */
+class Kimai_Api
+{
+
+    /**
+     * The API Key
+     *
+     * @var string
+     */
+    protected $apiKey;
+
+    /**
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


libraries/Kimai/Api.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @param array $params
+     */
+    public function __construct($params = array())
+    {
+        if (!is_array($params)) {
+            throw new \InvalidArgumentException('The configuration options must be an array.');
+        }
+        if (!array_key_exists('key', $params) || empty($params['key'])) {
+            throw new \InvalidArgumentException('Kimai API key is required.');
```
simonschaufi, line 58: my idea is that if apiKey is not provided, optional username/pw auth should be done instead like in Kimai_Remote_Api

kevinpapst:
What is your goal with this PR? 
Do you want to be able to build stuff like cli scripts or use Kimai in other PHP applications? 
Due to the limited dev resources, it would be better to improve the one API we have, instead of building another one.
Also, its a year old, do you still want to work on it, or shall we close it for now, and you keep your branch in your repo for possible future work?

simonschaufi:
I wanted to build another app based on kimai back then but then the whole project went into another direction, so I stopped working on it. The positive thing would have been, that there are no HTTP Requests in the background but instant access to the data. Anyway, I don't need that API any more.

PR #875, opened 2016-12-07
Added new functionality to allow LaTeX generation of invoices.

neerdoc:
- Added functionality for LaTeX invoices.

Reason for this pull request:
- LaTeX is one of the most complete type setting tools there are with huge community support.


extensions/ki_invoice/init.php
```diff
@@ -55,13 +55,29 @@
...
     if (!is_dir($tplFile)) {
         $extension = strtoupper($tplInfo['extension']);
+    } else {
+        //Check if index.hmtl or invoice.tex is there!
+        if (file_exists($tplFile.'/index.html')) {
+            $extension = 'HTML';
+        } elseif (file_exists($tplFile.'/invoice.tex')) {
+            $extension = 'LaTeX';
+            //Test if we can execute pdflatex
+            $output = exec($kga['LaTeXExec']);
```
kevinpapst: why don't you use `is_executable($kga['LaTeXExec'])`?

neerdoc: Didn't know about that function. I can use that instead if you like.

kevinpapst: Yes please.

kevinpapst:
Hi @neerdragon!
First of all: Thanks a lot for contributing such an interesting feature :+1:  I will test it out within the next days.

One thing I would request is check(s) for the existence of the `exec` function, as it is deactivated in many environments (shared hosts for example) for security reasons:
```
if(function_exists('exec')) {
    // deactivate latex functionality
}
```
I would suggest that we dislay an error message instead of the input field for the executable in the admin section and then disable it in the invoice section as well.

neerdoc:
Hi @kevinpapst !

Ok, I will do that! It is kind of automatically deactivated in the invoice section if it cannot execute, but I'll look into changing that. What about the error message? Is it ok to simply replace the input field with a message? Or should it be more "fancy", like a floater?

One more thing, I added a field in the database to store the executable, i.e. LaTeXExec, but I didn't dare change anything in the update database scripts... So I guess I need help with that if this is approved!

neerdoc:
Added fixes as suggested.

simonschaufi:
please no floater. otherwise it would open every time you go to admin panel which would be very annoying.

For the db migration: make sure the installer does contain the column and the updater adds it (and increment the version number!)

extensions/ki_adminpanel/init.php
```diff
@@ -121,6 +121,14 @@
         $view->assign('roundSeconds', $kga['conf']['roundSeconds']);
     }
 
+    if(!function_exists('exec')) {
+        // deactivate latex functionality
+        Kimai_Logger::logfile("Cannot execute external files. LaTeX invoices will be disabled.");
+        $view->assign('canExecute', false);
```
simonschaufi: I would name it "execAvailable". What do you think? canExecute sounds more like some internal stuff (object) can execute something.

simonschaufi (changes requested):

language/en.php
```diff
@@ -257,6 +257,8 @@
     "roundTimesheetEntries" => "Round new timesheet entries to",
     "minutes" => "Minutes",
     "seconds" => "Seconds",
+    "LaTeXExecutable" => "Full path to LaTeX executable, i.e., 'pdflatex'.",
+    "cannotExecute" => "Your system does not allow execution of external files. LaTeX invoice generation will be disabled.",
```
simonschaufi: Maybe a link where latex can be downloaded from?

neerdoc:
No floater. I just changed the input field to a text instead.

language/en.php
neerdoc: Sure, but then it needs to be updated in case the link changes... I believe anyone who wants to use LaTeX will know where to find it. If not, a simple "google LaTeX" will give you the answer... :)

extensions/ki_adminpanel/init.php
neerdoc: Sure. I can change it!

neerdoc: Fixed.

simonschaufi:
Where is the class Kimai_Invoice_LaTeXRenderer actually? did you forget to add it?

neerdoc:
Yes, indeed I forgot to add two files... Should be all now! :/

simonschaufi:
please adjust your code to make it PSR2 compatible (StyleCI check is green). click the Details link to automatically get a patch file and apply it.

simonschaufi:
I had a deeper look into your code now, made some changes and improvements but here are some things to consider:

* everything inside the libraries folder must be a class! Make OCR a class!
* did you write the OCR completely by your own? if not, sources?
* don't kill the hard drive! use as little as possible file IO.
* why do you execute pdflatex twice???
* please use single quotes in "normal strings"

simonschaufi (approved):

neerdoc:
Ok, I can change OCR to be a class.
Yes, I wrote the OCR function myself.
The LaTeX typesetting engine works that way. You always have to run it twice.

simonschaufi:
You somehow messed up the indentation. Could you please align your code properly? If you are using phpstorm, use the autoformating, that works pretty well :D

Please also make your db changes so that this PR can be merged.

libraries/Kimai/Invoice/LaTeXRenderer.php
```diff
@@ -0,0 +1,157 @@
...
+        // fetch variables from model to get values
+        $customer = $this->getModel()->getCustomer();
+        $projects = $this->getModel()->getProjects();
+        $entries = $this->getModel()->getEntries();
+        $data = $this->getModel()->toArray();
+
+        //Create invoiceId
+        $in = time();
+        $invoiceID = date('y', $in) . $customer['customerID'] . date('m', $in) . date('d', $in);
+        require_once 'Checksum.php';
```
simonschaufi: can be dropped as autoloading works out of the box if the class is correctly named.

libraries/Kimai/Invoice/LaTeXRenderer.php
```diff
@@ -0,0 +1,157 @@
...
+        }
+        //Run pdflatex again, throw error if not!
+        $output = exec('cd ' . $tempDir . ' && ' . $kga['LaTeXExec'] . ' ' . $data['invoiceId'] . '.tex');
+
+        //Return the rendered file
+        $this->sendResponse($tempDir . '/' . $data['invoiceId'] . '.pdf');
+    }
+
+    /**
+     * @return pdf
```
simonschaufi: does not return anything but just output the file

simonschaufi (changes requested):

libraries/Kimai/Invoice/LaTeXRenderer.php
neerdoc: Changed to "@return null".

libraries/Kimai/Invoice/LaTeXRenderer.php
neerdoc: Removed.

neerdoc:
@simonschaufi Please double check my change to installer.php and updater.php as I feel a bit unsure that I got it right.

installer/install.php
```diff
@@ -345,6 +345,7 @@ function quoteForSql($input) {
 ('roundSeconds', '0'),
 ('allowRoundDown', '0'),
 ('defaultStatusID', '1')
```
simonschaufi: comma missing in the end

simonschaufi (changes requested):
looks good so far but you forgot to raise the revision number in includes/version.php.

installer/install.php
neerdoc: Oops. 
Fixed. And increased revision by one in version.php

neerdoc:
@simonschaufi It's all good now?

simonschaufi:
I would like to test it myself before I merge it but didn't have the time yet.

neerdoc:
@simonschaufi Any progress on this issue?

simonschaufi:
Thank you for your ping!
Could you please also write some documentation how to setup a system (for all the noobs who don't know how to setup latex)? Please do that in the documentation repository.

Just for me, is this the minimal setup for ubuntu?
```
sudo apt-get install texlive texlive-lang-german texlive-latex-extra
```

simonschaufi (approved):

neerdoc:
Yes, I can add some documentation!
I'm running on FreeBSD, don't know about Ubuntu... But it should be enough to install any texlive distribution.

AaronFeldman:
I would love to test this out on ubuntu 17.10!!  Also, I can help proof the documentation if needed.  On Ubuntu it has (at least in the past) been a bad idea to use apt-get to install texlive, because the ubuntu packages are not updated often enough, at times years behind the current distribution. The best way is to use the net installer, as described here: [https://tex.stackexchange.com/a/1094/90087](https://tex.stackexchange.com/a/1094/90087)

dwagenk:
Hey,

this is the only PR or issue related to LaTeX for templates. But it looks like it's discontinued. Any chance for merging?

I'm evaluating kimai for tracking the time I spend on various freelance projects. I've got my invoices, letter-templates and basically everything else in LaTeX, so autogenerating those from within kimai would be great!

I'm not a php/web developer, but experienced enough with self hosting things like this, so I could test changes.
Got a Debian server available, but would usally configure and spin it up inside docker containers.

Thanks in advance!
Daniel

simonschaufi:
@DWagenk There are some conflicts with this branch as you can see so I can't just merge it like that. You can also checkout the source of the fork and see if that works for you. I don't want to waist too much time with Latex as you are the second person who is interested in it.

PR #880, opened 2017-01-02
Update en.php

neilzone:
Fix typo


simonschaufi:
ups sorry, wrong branch. actually all new pull requests should go into develop...

PR #889, opened 2017-01-13
Update / Change excel export from XLS to XLSX

jst2017:
- The Excel- Export should be possible in XLS and XLSX or only in XLSX

Reason for this pull request:
- In Kimai- Version 09.3 the only way to export to excel is to use the format XLS. Necessary is also a possibilty to export in format XLSX.

If an export to excel format XSLX is possible in later Kimai- Modules, is it possible to update only the export module in Kimai 09.3 to realize the Excel- Export in format XSLX?
If this is possible, how can this be done?


atlem:
mabe make this write directly from db to exel file in root folder om every new event. this way can have live preview and export live from sheet to browser. also able to use formate code from that sheet to another to custom time sheets or whatever .. kimai->db->kimai.exel->custom-timesheet.exel .

simonschaufi:
this PR is totally messed up.

PR #920, opened 2017-02-14
Implementation of GUI changes necessary to manage the authenticators.…

timlegge:
…  Does not currently do anything other than display the required fields for each authenticator

Adds the GUI to allow the authenticator to be managed via the webapp.  It does not do anything at present other than display fields for each authenticator

Reason for this pull request:
Mostly an FYI that I have something here.  I am working on it and will continue to allow the GUI actions to manage the authenticators.  It is not really ready to be merged I just wanted you to be aware of it.


timlegge:
This has the changes necessary to save and maitain values for the authenticators to the database.  Teh authenticators do not guttent use the values.  That is the next step.

It includes database changes and increments the version of the database to trigger the update

simonschaufi:
I don't have a good feeling storing external authentication data in the database. I would prefer storing them in the config file. Do you have a good reason it's better to have them in the db?

timlegge:
Not really a good reason.  I needed a place to store them and the
configuration table seemed good.  However, it caused me an issue with the
bindDN and bindPW that the database does not want to accept an empty string
for since it converts it to a  NULL that I had not resolved yet.

The only reason I guess was my plan to centralize all the authenticators
config values in one location.

by config file are you thinking about the autoconf file?

Tim

timlegge:
The only other reason I suppose would be to ease moving your installation between servers.  If all the config is in the database it allows you to restore the database on another server and not to have to figure out which config files you are missing.  I am good either way I guess but I will need to figure out the best way to save the settings if not in the configuration table

simonschaufi:
Yes, I'm thinking about the autoconf file.
In my opinion there should be only settings in the database that are used when you are already logged in.
This is how other systems also work. They also store the auth stuff in files.

timlegge:
Hi

As someone who supports a lot of systems I would likely lean to having as much as possible in the database.  Obviously though there needs to be a file with settings to store things like the database name, user and password, etc and could easily hold these settings.

That being said, you know this project and you have your preference so I will defer to you.  As long as those settings are brought together in one place I guess it does not matter.  Should I rewite the entire file if the config of one of the settings changes or implement a method of updating individual settings in the file.  Not having to work with files is a +1 for saving in the database :-)

Tim

simonschaufi:
I started an internal branch where I write the config file as array which is a lot easier to handle then single variables. Let me finish it asap and then you can use my API to write your values to the config file, ok?

timlegge:
Sounds fine.  I am continuing to get it to work as is but it should be a
simple matter of changing how it sets and gets the values.

Tim

timlegge:
I have pushed the final changes that appear to make the authenticators fully managable from the GUI and I have tested each of the authentication methonds (except the last ActiveDirectory change).

Included in the changes are several fixes to ensure that the default group gets correctly added to the User as they are created.  I tried to keep those in seperate commits but I may have included one in one of the other commits.

I am done what I wanted to do but when you get the code written to use a file to store it instead of the database I don't mind taking a look and retrofitting the changes .

Tim

timlegge:
Pulling this back to allow merge of issue 933 to be fixed

timlegge:
@simonschaufi Did you get time to look at the config file as an array that you mentioned.  I'd like to finish of the authenticator GUI work and based on the SMTP stuff I did I see how to write to the file.  I can change this GUI code to write to the autoconf.conf file in the current method but if you would prefer to wait till you are able release those changes before I change this code I will move back to generalizing the SMTP work a little more.  

In addition, I am thinking of adding SAML support for authentication as we are using it at work so I have a few "non core" features I can work on while I wait.

PR #948, opened 2017-04-06
Ldapadvanced: allow for removing prefix and suffix from user provided…

### LINKED ISSUE #946: Ldapadvanced bugs and user logins
#### Expected behaviour
Users login and sometimes provide a username in the wrong format:
example\username
username@example.com

the ldap authenticator handles it by stripping the prefix and suffix.  Ldapadvanced does not

also special characters in the distinguished name are not escaped and will prevent login

#### Actual behaviour
User cannot login in eithe case of wrong format or special characters
#### Steps to reproduce
Setup ldap and pass the username in the wrong format 
Have a dn that includes special characters

#### Reproducable on demo.kimai.org? [yes|no]
It would be with ldapadvanced against a open ldap or active directory server

#### Kimai version [for example: dev 1.1.0.1389|github master|github develop]
Kimai 1.1.0.1389 
#### Relevant log entries from log file (debug tab or temporary/logfile.txt)

For the special characters:
E_WARNING ldap_search(): Search: Bad search filter @/var/www/kimai_1.1.0/libraries/Kimai/Auth/Ldapadvanced.php line 333

I have preliminary fixes that I need to clean up that work in our environment.  I will send a pull request with better implementations for this:

        $prefix = 'example\\';
        $suffix = '@example.com';

       if (substr(strtoupper($username), 0, strlen($prefix)) == strtoupper($prefix)) {
           $username = substr($username, strlen($prefix));
       }
       if (substr(strtoupper($username), strlen($username)-strlen($suffix), strlen($username)) == strtoupper($suffix)) {
           $username = substr($username, 0, strlen($username) - strlen($suffix));
       }

and:

        $distinguishedName = str_replace('(' , '\28' , $distinguishedName );
        $distinguishedName = str_replace(')' , '\29' , $distinguishedName );

timlegge:
… username to reduce login errors

FIXES #946 

Remove a defined user prefix and suffix from the user provided username to reduce login issues

Reason for this pull request:
Add changes to fix up some Ldapadvanced issues


libraries/Kimai/Auth/Ldapadvanced.php
```diff
@@ -199,6 +213,17 @@ public function __construct($database = null, $kga = null)
      */
     public function authenticate($username, $password, &$userId)
     {
+        // Remove the prefix and suffix from the user name
+        $lenprefix = strlen($this->removeUsernamePrefix);
+        if (substr(strtoupper($username), 0, $lenprefix) == strtoupper($this->removeUsernamePrefix)) {
```
kevinpapst, line 218: this might work, but wouldn't it be easier to read:
```
if (strpos($username, $this->removeUsernamePrefix) === 0) { 
  $username = str_replace($this->removeUsernamePrefix, '', $username);
}
```
Something like that should work for the suffix as well

kevinpapst: Also: could you please do a new PR for the changes? merging so many commits for such a small change might be a bit too much ;-)

simonschaufi: >  could you please do a new PR for the changes?

please not. We will still squash them together in the end to make it just one.

kevinpapst:

PR #958, opened 2017-04-15
Bulgarian language

dobrevg:
- f3c0a9d: Adds singe file bg.php with almost full Bulgarian translation.
- 12cf96d: Changes on the Installer. Adds Bulgarian language to the kimai installer.

Reason for this pull request:
-  Add Bulgarian support


simonschaufi:
I would recommend changing the language checks to: check for other languages else use english as fallback.
Even better would be to use the translation which already exists in the normal app storing the translation strings in the language files.

kevinpapst:
Hey @dobrevg, the bg file for kimai itself is great!  Thanks :)
The installer needs some tweaks, as Simon said. Actually we should extract all translations from all files and put them in language files, but for now its fine if "en" is always the last else case.

simonschaufi (approved):

Issue #960, opened 2017-04-22
Proof of concept SAML authenticator


timlegge:
Hi

I managed to get a proof of concept of SAML authentication to work.  It is too early for  a pull request but I will provide code and information attached:

Basically:
1) Download OneLogin's SAML PHP Toolkit v2.10.5 from: https://github.com/onelogin/php-saml/releases and extract it to libraries/php-saml-2.10.5
2) apply the saml.diff patch
3) save Saml.php to libraries/Kimai/Auth
4) modify the includes/autoconf.php file to set the authenticator to 'saml'
5) Modify the libraries/Kimai/Auth/Saml.php to for your SAML provider
6) update the fingerprint of the public key from Google or whoever in  libraries/Kimai/Auth/Saml.php

I used the SAML Authenitaction in GSuite since I have a Google Apps account.  See the attached screen shots for the google required config

![kimai_google_attributes](https://cloud.githubusercontent.com/assets/1029883/25300452/28a810f6-26e6-11e7-883e-d7fbb7201632.png)
![kimai_google_saml](https://cloud.githubusercontent.com/assets/1029883/25300453/28a9a164-26e6-11e7-83a2-518b841a0080.png)

Remove the .txt extension
[Saml.php.txt](https://github.com/kimai/kimai/files/948664/Saml.php.txt)
[saml.diff.txt](https://github.com/kimai/kimai/files/948663/saml.diff.txt)

So some questions. 

1) php-saml seems to support composer and I see kimai uses it.  How does that work?
2) I assume that if composer is used it takes care of loading the library?

I have a fair amount of clean up and fixups todo as well as looking into ensure I have all the security recommendations enabled but I would welcome feed back.  I plan to test it against a Microsoft ADFS service as that is my target,

Tim

timlegge:
Trying to think of the best way to handle admin (i.e.) local logins in the case of SAML.  If saml is used for authentication I have it set to automatically redirect on accessing the index.php page.  Right now accessing: https://kimai.local/index.php?a=logout allows you to login an a local admin which is likely fine.

Thoughts?

Tim

PR #971, opened 2017-04-29
Saml auth2

### LINKED ISSUE #961: SAML support
Hi

I am looking at implementing SAML support for my use in our instalation.  I am now looking at how the php-saml library should be installed.

It looks like I can use composer but that would install its own vendor directory instead of the libraries directory that the other libraries use.  

Should I simply download the php-saml library and add it to the libraries directory or does it make sens to go the composer route?

Tim

timlegge:
FIXES #961 #960 

- 
Addition of SAML Authenticator.  Allows you to authenticate to Kimai via SAML.  Tested against Google's GSuite SAML Identity Provider

Reason for this pull request:
- 
More eyes on the code.  Anyone wanting to test contact me and I can provide a temporary account on my GSuite domain.

Documentation:
- 
https://github.com/kimai/documentation/pull/10


simonschaufi:
Thank you Tim!
Since I want to require composer sooner or later anyway for git installer, it will be better to not add so many library files into the core of kimai. For the releases it will be included but that is just a zip. If we add it now, it will always be in the history, even if we will delete them later on.
If everything is working properly now, I will merge it as soon as the library code is removed.
You can just delete everything and then amend the commit and force push.

timlegge:
I will work on it.  Just on vacation at the moment so it might be a week or two

philipp-doblhofer:
@timlegge Can you predict when it's ready to merge, because it would be a nice feature for kimai ;)

timlegge:
the changes are trivial.  I am rebasing and testing.  some time this week or weeken

simonschaufi:
@timlegge Hi Tim, could you please now remove all the composer dependency files and look for the conflicting files? I would be glad if I can merge this finally.

kevinpapst:

PR #978, opened 2017-05-09
Fix array init for PHP 7.1

FriedrichFroebel:
Fixes empty logfile output inside debug extension

- Init the `$filearray` variable as an array instead of a string.

Reason for this pull request:
- Seems like some array handling has been changed in PHP 7.1 leading to an empty logfile output inside the debug extension.


simonschaufi (approved):

simonschaufi:
Thank you!

PR #1028, opened 2017-08-18
[BUGFIX] show description header also if tracking numbers are not activated

peterwimmer:
FIXES #

- fix wrong table header position for description

Reason for this pull request:
- the position of the header had to be outside of tracking number check


simonschaufi (changes requested):
If you move the description outside of the check, the column itself also has to move outside in line 40.

simonschaufi (approved):

peterwimmer:
missed it.... it is changed now

PR #1035, opened 2017-08-25
TASK: Lots of code cleanup

simonschaufi:
- include `$kga = Kimai_Registry::getConfig();` where needed
- include `$database = Kimai_Registry::getDatabase();` where needed


core/kimai.php
```diff
@@ -19,11 +19,13 @@
...
 // = Smarty (initialize class) =
-// ============================= 
+// =============================
 require_once '../includes/basics.php';
 
+$database = Kimai_Registry::getDatabase();
+
 $view = new Zend_View();
-$view->setBasePath(WEBROOT . '/templates');
+$view->setBasePath(WEBROOT . 'templates');
```
kevinpapst: Here and for all following occurences:
- the constant should always end with NO trailing slash
- the usages should always incorporate a starting slash

core/processor.php
```diff
@@ -278,7 +280,8 @@
                     $errorMessages['customerGroups'] = $kga['lang']['atLeastOneGroup'];
                 }
 
-                if (!checkGroupedObjectPermission('Customer', $id ? 'edit' : 'add', $oldGroups, $_REQUEST['customerGroups'])) {
+                if (!checkGroupedObjectPermission('Customer', $id ? 'edit' : 'add', $oldGroups,
+                    $_REQUEST['customerGroups'])) {
```
kevinpapst, line 284: Why the break? what about increasing the line length limit or simply ignore some long lines?
Just for me, this is less readable than before - and before it took only half of my screen...
Same goes for all following places.

extensions/ki_expenses/processor.php
```diff
@@ -217,8 +220,10 @@ function ($sum, $expense) {
                 $kga['lang']['errorMessages']['emptyField'],
                 $kga['lang']['designation']
             );
-        } else if (!is_numeric($_REQUEST['edit_value'])) {
-            $errors['edit_value'] = $kga['lang']['errorMessages']['wrongData'];
+        } else {
+            if (!is_numeric($_REQUEST['edit_value'])) {
```
kevinpapst: why split it off in a separate if?

kevinpapst (commented):
To be honest, I stopped after 10 minutes, this is way to large to review it properly. Besides, it looks like automatic code formatting, which is bitchy with so many oen PR - you might create a lot of conflicts ;-)
Are all these code formattings part of the scrutinizer check?

simonschaufi:
>  Are all these code formattings part of the scrutinizer check?

no, i did everything by hand.

core/kimai.php
simonschaufi: done

core/processor.php
simonschaufi: done

extensions/ki_expenses/processor.php
simonschaufi: misunderstanding.

PR #1036, opened 2017-08-26
[FEATURE] Billability can now be changed in the timesheet list directly

peterwimmer:
- add an option for changing billability directly in the timesheet list

Reason for this pull request:
- This change speeds up our kimai handling a lot. We can now change the billability directly in the list view. We do this a lot in ongoing longer projects.

### DIFF
--- a/extensions/ki_timesheets/init.php
+++ b/extensions/ki_timesheets/init.php
@@ -111,14 +111,16 @@ $view->assign('hideComments', true);
 $view->assign('showOverlapLines', false);
 $view->assign('showTrackingNumber', false);
 $view->assign('showBillability', false);
+$showBillability = false;
 
 if (isset($kga['user'])) {
     $view->assign('hideComments', !$kga->getSettings()->isShowComments());
     $view->assign('showOverlapLines', $kga->getSettings()->isShowOverlapLines());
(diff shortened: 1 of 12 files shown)


extensions/ki_timesheets/processor.php
```diff
@@ -433,12 +464,14 @@ function($activity) {
         $view->assign('hideComments', true);
         $view->assign('showOverlapLines', false);
         $view->assign('showTrackingNumber', false);
+        $view->assign('showBillability', false);
```
simonschaufi: I would split this into a variable and assign it only once after the block below.

extensions/ki_timesheets/init.php
```diff
@@ -110,11 +110,13 @@
 $view->assign('hideComments', true);
 $view->assign('showOverlapLines', false);
 $view->assign('showTrackingNumber', false);
+$view->assign('showBillability', false);
```
simonschaufi: same here

simonschaufi (changes requested):

extensions/ki_timesheets/init.php
peterwimmer: done :)

simonschaufi (approved):

PR #1037, opened 2017-08-26
[FEATURE] preset end day to the value of start day if other date/time…

peterwimmer:
… values are changed and the end day was not set before

- prefill end date in timesheet floater on change of other time/date values. The prefill value will be the value of the start day. The end day will only be changed if there is not already a value inserted.

Reason for this pull request:
- This is also speed improvement for us. While adding entries e.g. from handwritten lists it is more convenient when the end day will be set if you change/add the end time. This improves/speeds up this task quite a bit.

### DIFF
--- a/extensions/ki_timesheets/templates/scripts/floaters/add_edit_timeSheetEntry.php
+++ b/extensions/ki_timesheets/templates/scripts/floaters/add_edit_timeSheetEntry.php
@@ -84,7 +84,7 @@ $autoSelection = $this->kga->getSettings()->isUseAutoSelection();
                     <li>
                         <label for="start_day"><?php echo $this->kga['lang']['day'] ?>:</label>
                         <input id='start_day' type='text' name='start_day' value='<?php echo $this->escape($this->start_day) ?>' maxlength='10' size='10' tabindex='6'
-                               onChange="ts_prefillEndDate(); ts_timeToDuration();" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
+                               onChange="ts_timeToDuration();" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
                         -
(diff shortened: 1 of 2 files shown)


extensions/ki_timesheets/templates/scripts/floaters/add_edit_timeSheetEntry.php
```diff
@@ -84,7 +84,7 @@
                     <li>
                         <label for="start_day"><?php echo $this->kga['lang']['day'] ?>:</label>
                         <input id='start_day' type='text' name='start_day' value='<?php echo $this->escape($this->start_day) ?>' maxlength='10' size='10' tabindex='6'
-                               onChange="ts_timeToDuration();" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
+                               onChange="ts_prefillEndDate(); ts_timeToDuration();" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
```
simonschaufi: this line does not work for me. If I only change the start day, the end day field is still empty. Could you please figure out why?
If I change the time, then the end day field is filled.

simonschaufi (changes requested):

extensions/ki_timesheets/templates/scripts/floaters/add_edit_timeSheetEntry.php
peterwimmer: Yes, it seems the datepicker blocks this. Here i think we can remove the call anyway. In the longer run i will get back to this (still working on the report feature which i want to get finished soon :))

simonschaufi (approved):

extensions/ki_timesheets/templates/scripts/floaters/add_edit_timeSheetEntry.php
simonschaufi: would still be interesting why it doesn't work there.

PR #1043, opened 2017-08-30
Add Photos

carrolltondatago:
FIXES #

- add a way to attach photos to a job,  before, at completion of a task and after.  Then attach the before and after to the invoice.

Reason for this pull request:
- What we do is very visual and customers like to see how the job is going.


simonschaufi:
I have no idea what you have done but this pull request is not including any of your commits. Please also merge into develop and not master!

PR #1045, opened 2017-08-31
BUGFIX #905

### LINKED ISSUE #905: GUI sync issues with Start/Stop button.
Expected behaviour:

Actual behaviour:

Steps to reproduce:

Kimai version Stable 1.1.0

RStudios:
FIXES #905

Reason for this pull request:
- Shouldn't reset the global 'offset' variable that calculates difference between server time and local time
- Fixes issues described in https://github.com/kimai/kimai/issues/905
- Awaits 'updateRecordStatus' before re-enabling buzzer

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -328,11 +328,14 @@ function startRecord(projectID,activityID,userID) {
     startsec = now;
     show_stopwatch();
     value = projectID +"|"+ activityID;
+	$('#buzzer').addClass('disabled');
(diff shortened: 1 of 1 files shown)


RStudios:
Sorry, pushed to wrong branch, can you change it to kimai:develop

simonschaufi:
I can't change it. Github does not have this feature. You need to make a new pull request.

PR #1047, opened 2017-08-31
BUGFIX #905

### LINKED ISSUE #905: GUI sync issues with Start/Stop button.

RStudios:
FIXES #905

Reason for this pull request:

Shouldn't reset the global 'offset' variable that calculates difference between server time and local time
Fixes issues described in #905
Awaits 'updateRecordStatus' before re-enabling buzzer @

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -256,10 +256,10 @@ function n_uhr() {
 // grabs entered timeframe and writes it to database
 // after that it reloads all tables
 //
-function setTimeframe(fromDate,toDate) {
+function setTimeframe(fromDate,toDate, callback) {
 
     timeframe = '';
-
+    
     if (fromDate != undefined) {
       setTimeframeStart(fromDate);
       timeframe += strftime('%m-%d-%Y',fromDate);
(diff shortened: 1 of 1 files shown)


RStudios:
Third attempt to do it right, should be in correct branch now

RStudios:
Added extra commit code:

Fixed calculation of the offset value to prevent negative timing
Disabled the buzzer while awaiting start/stop response from server
Corrected race condition and refresh of table when starting a new timer
Updated setTimeframe to take an optional callback parameter instead of the default 'hook_timeframe_changed'

simonschaufi:
🎉

simonschaufi:
@RStudios I noticed one more issue since I'm testing this since some weeks now. When you start the timer and then modify the currently running record (give it an earlier starting time), the timer on the right top doesn't update to the new time. Can you confirm that?

RStudios:
@simonschaufi I'll look into it

simonschaufi:
Another thing which was working before: Whenever you reaload the page, the couter starts from 0 again. This was working before as I was reading the start time from the server and then continued counting. This has to work so that I can finally merge it into master.

kevinpapst:
what about reverting this PR and postpone its integration for the next release?

simonschaufi:
@RStudios Today I made a new release but excluded this pull request from it as it is still not fixed completely. Could you please have a look into it so i can make a new release with this great bugfix included?

PR #1096, opened 2018-01-24
Update customers.php

DunklesKaltesNichts:
Reason for this pull request:

i haven't seen any customers. it just showed "no Entries".
i looked at the source code and looked for the bug and i think i found it.

if i change this comparison, i see customers again.


simonschaufi:
thank you for fixing this!

PR #1113, opened 2018-02-15
Update html.php

IxFail:
Removed unnecessary @ line 213 with would show as DDescription at export instead of Description

FIXES #

- Removed unnecessary D @ line 213

Reason for this pull request:
- 
Export would show unnecessary D at Description @ line 213


simonschaufi (approved):
makes total sense. I wonder where this comes from...

PR #1132, opened 2018-02-23
[FEATURE] tinyDoc-->openTBS; support for .docx templates

AlxBe:
Added openTBS support for supporting MS Word invoice templates. Due to
the changes the naming of the blocks in the templates had to be changed.
Changed generation of 'invoiceId' in ki_invoice/print.php (german regulation needs increasing invoice numbers). Invoice filename is generated starting with invoiceId.
Added support for numbering positions of a invoice with tag
[pos.position] in invoice templates (see template 2018)
All .odt templates must be migrated:
1. replace all [var.XXXX] tags to [kimai.XXXX]
2. replace all [row.XXXX] tags to [pos.XXXX]
3. row tags have to be changed to [...block=tbs:row..]
Use of .doc(x) templates is similar to .odt.

All odt templates are migrated to openTBS.

FIXES #

Uses openTBS as library (maintained)

Reason for this pull request:
Use MS Word for templates. Process simplification


kevinpapst:
I didn't check the reason yet: why do we have to adapt existing odt templates?
This is a major backward compatibility break. We cannot merge it if there is no fix for that.
Imagine: people update Kimai and all of a sudden all Kimai of users cannot create invoices...

AlxBe:
The reason for changing the [var.XXXX] tags to [kimai.XXXX] is, that openTBS seams to refer to global variables if this tag is used. I did't spend much time to investigate this in openTBS. To get a true migration, I wanted to remove global vars and tinyDoc kompleately. Possible solution to be backwards compatible are:
* create a script which migrates existing templates (include it in docxOdtRenderer.php)
* include a switch or a detection mechanism witch automatically converts the templates before using it (does not touch the original templates)
* do not touch the existing implementation of odtRenderer.php (keep tinyDoc)
*  do it the hard way: all users have to replace their template (it is just a simple replace of tags)
* keep openTBS using global vars (to be tested)

AlxBe:
Further on, it makes sense, to migrate the html templates to openTBS too. Of course with the same implications.

PR #1133, opened 2018-02-23
Added default value for billable (autoconfig.php)

simonschaufi:
default value for billable can be choosen in autoconfig.php. Main
advantage is, that you can select the billable valule, which should be
preset.

### DIFF
--- a/includes/basics.php
+++ b/includes/basics.php
@@ -26,10 +26,6 @@
 defined('WEBROOT') || define('WEBROOT', dirname(dirname(__FILE__)) . DIRECTORY_SEPARATOR);
 defined('APPLICATION_PATH') || define('APPLICATION_PATH', realpath(dirname(__FILE__) . '/../'));
 
-if (!file_exists(WEBROOT . 'libraries/autoload.php')) {
-    die('Please run <code>composer install --no-dev</code> on the command line to install all php dependencies.');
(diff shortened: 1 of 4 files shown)


includes/basics.php
```diff
@@ -26,10 +26,6 @@
 defined('WEBROOT') || define('WEBROOT', dirname(dirname(__FILE__)) . DIRECTORY_SEPARATOR);
 defined('APPLICATION_PATH') || define('APPLICATION_PATH', realpath(dirname(__FILE__) . '/../'));
 
-if (!file_exists(WEBROOT . 'libraries/autoload.php')) {
```
AlxBe, line 29: accidently deleted this one.... please restore

libraries/Kimai/Config.php
```diff
@@ -112,7 +116,15 @@ public function isTrackingNumberEnabled()
     {
         return $this->get('show_TrackingNr');
     }
-
+ /* Dublicate function
+    **
+     * @return array
```
AlxBe, line 121: Just delete - I don't know how to do it in this stage of the process....

kevinpapst:
Why so complicated? Just re-order the billable array and the first entry is preset. The autoconfig file was not meant to store arbitrary data.

PR #1134, opened 2018-02-24
Add SAML authentication with onelogin

### LINKED ISSUE #961: SAML support

timlegge:
FIXES #961 #960 

- 
Addition of SAML Authenticator.  Allows you to authenticate to Kimai via SAML.  Tested against Google's GSuite SAML Identity Provider. Supports MS ADFS and Google for Work authentication.

Reason for this pull request:
- 
More eyes on the code.  Anyone wanting to test contact me and I can provide a temporary account on my GSuite domain.

Documentation:
- 
https://github.com/kimai/documentation/pull/10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
         "tecnickcom/tcpdf": "^6.2.12",
         "zendframework/zendframework1": "^1.12",
         "phpoffice/phpexcel": "1.8.*",
-        "tinybutstrong/tinybutstrong": "^3.10"
+        "tinybutstrong/tinybutstrong": "^3.10",
+        "onelogin/php-saml": "^2.12"
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,61 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "096359b77016a53ca38310fe725a16aa",
+    "content-hash": "e148ed63ce9ab602f06c15fa7d023bf2",
     "packages": [
+        {
+            "name": "onelogin/php-saml",
+            "version": "v2.12.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/onelogin/php-saml.git",
+                "reference": "9416fa739cff88ffcc472967cec048a774b97e0f"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/onelogin/php-saml/zipball/9416fa739cff88ffcc472967cec048a774b97e0f",
+                "reference": "9416fa739cff88ffcc472967cec048a774b97e0f",
+                "shasum": ""
+            },
+            "require": {
+                "ext-curl": "*",
+                "ext-dom": "*",
+                "ext-mcrypt": "*",
+                "ext-openssl": "*",
+                "php": ">=5.3.2"
+            },
+            "require-dev": {
+                "pdepend/pdepend": "1.1.0",
+                "phploc/phploc": "*",
+                "phpunit/phpunit": "4.8",
+                "satooshi/php-coveralls": "1.0.1",
+                "sebastian/phpcpd": "*",
+                "squizlabs/php_codesniffer": "2.9.0"
+            },
+            "suggest": {
+                "ext-gettext": "Install gettext and php5-gettext libs to handle translations"
+            },
+            "type": "library",
+            "autoload": {
+                "classmap": [
+                    "extlib/xmlseclibs",
+                    "lib/Saml",
+                    "lib/Saml2"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "description": "OneLogin PHP SAML Toolkit",
+            "homepage": "https://developers.onelogin.com/saml/php",
+            "keywords": [
+                "SAML2",
+                "onelogin",
+                "saml"
+            ],
+            "time": "2017-11-06T17:44:03+00:00"
+        },
         {
             "name": "phpoffice/phpexcel",
             "version": "1.8.1",
--- a/index.php
+++ b/index.php
@@ -70,6 +70,9 @@ $justLoggedOut = false;
 if ($_REQUEST['a'] == 'logout') {
     setcookie('kimai_key', '0');
     setcookie('kimai_user', '0');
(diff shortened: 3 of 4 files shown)


simonschaufi:
I can only trust you that it works as I have no way to test and verify it.

simonschaufi (approved):

kevinpapst:
I will try to test that with my Google account. @timlegge could you share some insights how to use it?

timlegge:
@kevinpapst do you have a Google for organizations domain?  see https://github.com/kimai/documentation/blob/master/administrator/authenticator.md for the basics.

if you just have plain gmail I can create a  test account on my Google domain for you and @simonschaufi and send you the auth config to use

kevinpapst:
I have a Google for work account, is that sufficient?

timlegge:
@kevinpapst if you have the ability to manage the domain yes.  

1. Login to amnage the domain
2. In the admin select security
3. Set up single sign-on (SSO)
4. The settings are there 
5. Add the following items to includes/auth.php
    'saml_strict' => true,
    'saml_debug' => true,
    'saml_baseurl' => '',
    'saml_spentityId' => 'https://kimai-dev.com',
    'saml_spacsURL' => 'https://kimai-dev.com/index.php',
    'saml_spslsURL' => 'https://kimai-dev.com/index.php',
    'saml_spx509cert' => '',
    'saml_spprivateKey' => '',
    'saml_idpentityId' => 'https://accounts.google.com/o/saml2?idpid=????????????',
    'saml_idpssoURL' => 'https://accounts.google.com/o/saml2/idp?idpid=???????????',
    'saml_idpslsURL' => '',
    'saml_idpcertFingerprint' => '????????????????????????????????????',
    'saml_idpcertFingerprintAlgorithm' => 'sha1',
    'nonSAMLAcounts' => array('admin'),

Obviously update them with your settings.  I seem to remember having a better document around somewhere.

timlegge:
Also found this where I referenced the https://github.com/kimai/kimai/issues/960 attribut mapping that you need to do.

timlegge:
Also see https://support.google.com/a/answer/6087519?hl=en

The attribute mapping is important.  In the identity provider the following must be sent
ln = Last Name
fn = First Name
mail = Primary Email Address

timlegge:
@kevinpapst were you able to test?  I have now tested it with:

1. Google GSuite
2. Microsoft ADFS
3. Microsoft Azure AD

So far I have not run into any issues,

@simonschaufi  I have sent you a user id for my google for business domain so you can test.  I don't have Kevin's email address so I can'e easiy get him the account if he needs one.

Tim

simonschaufi:
awesome. I hope I will have time soon.

You will find kevin's email address in the git logs but ask him with which email address you should add him if that matters.

timlegge:
really only needed to send the invite to.  you will need to sign in with
the account and password details ypu setup there.  i sent you some details
for the kimai setup on your end.

timlegge:
@kevinpapst the account I sent you a few days can login at https://kimai.slyrain.com if you want to test it out.  I will probably have the site up for a few days.  Email me or Simon for the admin password

composer.json
```diff
@@ -15,7 +15,8 @@
         "tecnickcom/tcpdf": "^6.2.12",
         "zendframework/zendframework1": "^1.12",
         "phpoffice/phpexcel": "1.8.*",
-        "tinybutstrong/tinybutstrong": "^3.10"
+        "tinybutstrong/tinybutstrong": "^3.10",
+        "onelogin/php-saml": "^2.12"
```
simonschaufi, line 19: @kevinpapst shall we ship it with every release or put it in the "suggest" section so that when you need it, you should install it yourself?

kevinpapst: Interesting question. As not man people yet asked for saml, it might not be the most used feature, but then: how many people execute composer install and see the suggest messages (also, the list can get quite long as all suggestions are dumped and not only the ones from Kimai).
It doesn't add any package by itself but requires the following PHP modules:
```
        "php": ">=5.3.2",
        "ext-curl": "*",
        "ext-openssl": "*",
        "ext-dom": "*",
        "ext-mcrypt": "*"
```
I am not sure how good that works on Windows machines...
So probably adding it to suggest and clearly stating how to install it in the installation docu might be better? @timlegge what do you think?

timlegge: I guess either works.  I like the way composer pulls it in when the rest is done.

If I understand correctly, the tar file that is prepared for installation has composer install -nodeps run before it is prepared so the end user install would have the required libraries.  That is probably best because the install is ready for either method.

As people realize that saml is available via office365 and GSuite I suspect there would be more uptake because because it is a lot easier to configure than ldap.

Basically I would err to making it a simple change the authenticator and go (given I will resurect my GUI authentication changes after this is committed it will become easier for people to change authentication methods).

Tim

simonschaufi: The release contains `composer install --no-dev`. I have really no overview over our users how they normally authenticate but we get lots of ldap requests as you have noticed as well. You were the only one who talked about saml so far. I'm find to ship it even though the zip file will be a lot larger afterwards (~4MB uncompressed). If we exclude the docs from the zip, that would be fine for me.

timlegge: Someone asked about it on https://github.com/kimai/kimai/pull/971 which was the older PR that I need to close.  

Yes, the library does add some size for sure.  I am spoiled with 100Mb + fiber with unlimited bandwidth so it does not bother me,  I down load new Ubuntu cds because it is faster than finding where I but the last iso.

Putting it in avoids documenting how to install it so I am all for that...

simonschaufi: As of PHP 7.1, the php-mcrypt was deprecated. And as of PHP 7.2 it was completely removed. :( I'm really not sure what to do with this PR.

timlegge: It appears that they are working on support for 7.1 and the removal of the dependancy but not date is available at this time.

I am willing to work on it and update it at that time.

I do have a couple of questions though?

1. Is there still a concern regarding including the module in the repository because of the size?
2. What is the status of Kimai 2?  Should I be targeting that version?

Tim

simonschaufi: I don't mind adding it but I already hear people complaining that it doesn't work with PHP 7.2 as the module is not installed.
As Kimai 2 is completely new, the best is to ask Kevin. He is the main developer of it.

kevinpapst:

PR #1157, opened 2018-04-06
Bugfix: Clear expenses separately from timesheet entries

simonschaufi:
- Clear expenses separately from timesheet entries

Reason for this pull request:
- Clearing of entries was only possible when there were no expenses. With expenses resulted in a mysql error.

### DIFF
--- a/libraries/Kimai/Database/Mysql.php
+++ b/libraries/Kimai/Database/Mysql.php
@@ -2444,15 +2444,35 @@ class Kimai_Database_Mysql
      */
     public function setTimeEntriesAsCleared(array $entries)
     {
-        $ids = array_map(function ($entry) {
-            return $entry['timeEntryID'];
-        }, $entries);
+        // timesheet entries
+        $timeSheetEntries = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'timeSheet';
+        });
 
-        $update['cleared'] = 1;
+        $ids = array_map(function ($entry) {
+            return $this->conn->SQLFix($entry['timeEntryID']);
+        }, $timeSheetEntries);
+
+        $update = ['cleared' => 1];
 
         $where = ['timeEntryID IN (' . implode(',', $ids) . ')'];
 
-        return $this->conn->UpdateRows($this->getTimeSheetTable(), $update, $where);
+        $resultTimeSheet = $this->conn->UpdateRows($this->getTimeSheetTable(), $update, $where);
+
+        // expenses
+        $expenses = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'expense';
+        });
+
+        $ids = array_map(function ($entry) {
+            return $this->conn->SQLFix($entry['expenseID']);
+        }, $expenses);
+
+        $where = ['expenseID IN (' . implode(',', $ids) . ')'];
+
+        $resultExpenses = $this->conn->UpdateRows($this->getExpenseTable(), $update, $where);
+
+        return $resultTimeSheet && $resultExpenses;
     }
 
     /**
--- a/extensions/ki_invoice/private_func.php
+++ b/extensions/ki_invoice/private_func.php
@@ -83,6 +83,7 @@ function invoice_get_data($start, $end, $projects, $filter_cleared, $short_form)
         foreach($expenses as $entry) {
             $arr = ext_invoice_empty_entry();
 
+            $arr['expenseID'] = $entry['expenseID'];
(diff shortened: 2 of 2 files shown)


libraries/Kimai/Database/Mysql.php
```diff
@@ -2444,15 +2444,37 @@ public function save_timeframe($timeFrameBegin, $timeframeEnd, $user)
...
+        // expenses
+        $expenses = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'expense';
+        });
+
+        $ids = array_map(function ($entry) {
+            return $entry['expenseID'];
+        }, $expenses);
+
+        $update['cleared'] = 1;
```
kevinpapst: Is update changed in UpdateRows()? Otherwise this line seems not neccessary.

libraries/Kimai/Database/Mysql.php
```diff
@@ -2444,15 +2444,37 @@ public function save_timeframe($timeFrameBegin, $timeframeEnd, $user)
...
+        $timeSheetEntries = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'timeSheet';
+        });
+
         $ids = array_map(function ($entry) {
             return $entry['timeEntryID'];
-        }, $entries);
+        }, $timeSheetEntries);
 
         $update['cleared'] = 1;
```
kevinpapst: Is update defined before? Looks strange without an array initialization

kevinpapst:
Looks good beside the sql escaping

libraries/Kimai/Database/Mysql.php
simonschaufi: I can also write it like this: `$update = ['cleared' => 1]`. It's basically the same.

libraries/Kimai/Database/Mysql.php
simonschaufi: ok, i can leave out the second one.

simonschaufi:
> sql escaping

what do you mean?

kevinpapst:
I meant prevention of SQL injection attacks, e.g. using `Kimai_Database_Mysql::clean_data()`or likely even better `MySQL::SQLFix()`

libraries/Kimai/Database/Mysql.php
kevinpapst: afaik the current version would raise at least a note when using strict more in php7 as you assign a value to a non existing array.

simonschaufi:
Where should I use `MySQL::SQLFix()`?

libraries/Kimai/Database/Mysql.php
```diff
@@ -2444,15 +2444,35 @@ public function save_timeframe($timeFrameBegin, $timeframeEnd, $user)
...
-        return $this->conn->UpdateRows($this->getTimeSheetTable(), $update, $where);
+        $resultTimeSheet = $this->conn->UpdateRows($this->getTimeSheetTable(), $update, $where);
+
+        // expenses
+        $expenses = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'expense';
+        });
+
+        $ids = array_map(function ($entry) {
+            return $entry['expenseID'];
```
kevinpapst: You could do it here: `return MySQL::SQLFix($entry['expenseID']);`

libraries/Kimai/Database/Mysql.php
```diff
@@ -2444,15 +2444,35 @@ public function save_timeframe($timeFrameBegin, $timeframeEnd, $user)
      */
     public function setTimeEntriesAsCleared(array $entries)
     {
+        // timesheet entries
+        $timeSheetEntries = array_filter($entries, function ($entry) {
+            return $entry['type'] == 'timeSheet';
+        });
+
         $ids = array_map(function ($entry) {
             return $entry['timeEntryID'];
```
kevinpapst: You could do it here: `return MySQL::SQLFix($entry['timeEntryID']);`

simonschaufi: SQLFix is not static. I need to use `MySQL::SQLValue($entry['timeEntryID'], MySQL::SQLVALUE_NUMBER)`. ok?

kevinpapst: What about `$this->conn->SQLFix()` ?

simonschaufi: done.

kevinpapst (approved):
👍

PR #1181, opened 2018-05-18
Fixing time offset calculation on GUI

### LINKED ISSUE #905: GUI sync issues with Start/Stop button.

Viras-:
FIXES #905 / #1047

- Fixing time synchronisation code - this can only be done by combining server and client timestamp - not on the client only as proposed in the last pull

Reason for this pull request:
- Invalid offset calculation


simonschaufi:
Thank you for taking some time to fix this. unfortunatelly you added 2 commits. The best is to always create a new branch based on develop if you start working on a new issue. Are you good enough in git to remove the second commit?

Viras-:
I've created a new pull request with more clean code base - closing this one now.

PR #1192, opened 2018-05-25
Floater refactoring

simonschaufi:
Load the data from an array instead of assigning every single value to the view.

### DIFF
--- a/extensions/ki_expenses/floaters.php
+++ b/extensions/ki_expenses/floaters.php
@@ -26,29 +26,20 @@ $database = Kimai_Registry::getDatabase();
 require 'private_db_layer_mysql.php';
 
 switch ($axAction) {
-    case "add_edit_record":
+    case 'add_edit_record':
         if (isset($kga['customer'])) {
             die();
         }
 
-        $view->assign('projects', makeSelectBox("project", $kga['user']['groups'])); // select for projects
-        $view->assign('activities', makeSelectBox("activity", $kga['user']['groups'])); // select for activities
+        $projects = makeSelectBox('project', $kga['user']['groups']);
+        $view->assign('projects', $projects); // select for projects
+        $view->assign('activities', makeSelectBox('activity', $kga['user']['groups'])); // select for activities
 
         // ==============================================
         // = display edit dialog for timesheet record   =
         // ==============================================
         if ($id) {
             $expense = get_expense($id);
-            $view->assign('id', $id);
-            $view->assign('comment', $expense['comment']);
-            $view->assign('edit_day', date("d.m.Y", $expense['timestamp']));
-            $view->assign('edit_time', date("H:i:s", $expense['timestamp']));
-            $view->assign('multiplier', $expense['multiplier']);
-            $view->assign('edit_value', $expense['value']);
-            $view->assign('designation', $expense['designation']);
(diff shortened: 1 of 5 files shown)


kevinpapst:
What is the reason behind this PR? I don't like rewriting code that works when there is no real benefit from that rewrite afterwards. To often something breaks ... and thats not related to your work but a general rule of thumb that proves to be true every day in every team^^

simonschaufi:
I extended the table and don't need to assign every single variable to the view any more. It will be there automatically since the whole record is now available. this is good practice generally.

extensions/ki_expenses/floaters.php
```diff
@@ -62,19 +53,26 @@
                 $view->projects[$projectData['projectID']] = $customerData['name'] . ':' . $projectData['name'];
             }
         } else {
-            $view->assign('id', 0);
-            $view->assign('edit_day', date("d.m.Y"));
-            $view->assign('edit_time', date("H:i:s"));
-            $view->assign('multiplier', '1' . $kga['conf']['decimalSeparator'] . '0');
+            // defaults
+            $expense = [
```
kevinpapst, line 57: what about: move the defaults above if($id) in line 40 and then do a `$expense = array_merge($expense, get_expense($id))`?

extensions/ki_expenses/templates/scripts/floaters/add_edit_record.php
```diff
@@ -35,67 +35,67 @@
...
-                            ?>
-                            <br/>
+                            ?><br/>
                             <input type="text" style="width:395px;margin-top:3px" tabindex="2" size="10" name="filter" id="filter" onkeyup="filter_selects('add_edit_expense_project_ID', this.value);"/>
                         </div>
                     </li>
                     <li>
                         <label for="edit_day"><?php echo $this->translate('day') ?>:</label>
-                        <input id='edit_day' type='text' name='edit_day' value='<?php echo $this->escape($this->edit_day) ?>' maxlength="10" size="10" tabindex="5" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
+                        <input type="text" name="edit_day" id="edit_day" value='<?php echo $this->escape(date('d.m.Y', $this->expense['timestamp'])) ?>' maxlength="10" size="10" tabindex="5" <?php if ($autoSelection): ?> onclick="this.select();" <?php endif; ?> />
```
kevinpapst: can you re-use one of the existing date / time formats?

kevinpapst:
> this is good practice generally

says who?

extensions/ki_expenses/floaters.php
simonschaufi: I thought about it for a while in the beginning but then had the idea about completely separating the code into a "create" and an "update" Action. This is my end goal to make it even cleaner. I really don't like the mix of both together in one method.

extensions/ki_expenses/templates/scripts/floaters/add_edit_record.php
simonschaufi: sure!

simonschaufi:
The controller should not care about the view. The view should be able to display all properties of an object (array in our case).

kevinpapst:
Good argument!

kevinpapst (approved):

PR #1201, opened 2018-06-09
Api changes

xakiru:
## Changes proposed in this pull request:
- updateActiveRecording : added "comment" to the proprities that could be updated
- updateActiveRecording : added optional parameter timeEntryID to be able to edit/stop a specific recording 
- startRecord : returns the timeEntryID of the started recording instead of empty array.

## Reason for this pull request:
- I launched KimaiDesktop* interface that use the api to make start/stop recordings one click away.
- The fixes doesn't change the behaviours of the api (no previous application would break)
- These features are basic and would be needed by others anyway.

 * KimaiDesktop: I just finished the first version, I will need to cleanup code before posting it, but the binary is on the repo for testing https://github.com/xakiru/KimaiDesktop

### DIFF
--- a/libraries/Kimai/Remote/Api.php
+++ b/libraries/Kimai/Remote/Api.php
@@ -161,7 +161,7 @@ class Kimai_Remote_Api
 
         $result = $this->getBackend()->startRecorder($projectId, $activityId, $uid);
         if ($result) {
-            return $this->getSuccessResult([]);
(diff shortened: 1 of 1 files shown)


simonschaufi:
Thank you for improving the API. Unfortunatelly you used the wrong branch. 

Please read the guidelines on how to make code changes!

- First, we only develop on `develop` branch. 
- Second, always create a new branch based on develop and don't merge your develop branch into ours.

PR #1202, opened 2018-06-09
Remote Api Changes

xakiru:
## Changes proposed in this pull request:
- updateActiveRecording : added "comment" to the proprities that could be updated
- updateActiveRecording : added optional parameter timeEntryID to be able to edit/stop a specific recording 
- startRecord : returns the timeEntryID of the started recording instead of empty array.

## Reason for this pull request:
- I launched KimaiDesktop* interface that use the api to make start/stop recordings one click away.
- The fixes doesn't change the behaviours of the api (no previous application would break)
- These features are basic and would be needed by others anyway.

 * KimaiDesktop: I just finished the first version, I will need to cleanup code before posting it, but the binary is on the repo for testing https://github.com/xakiru/KimaiDesktop


libraries/Kimai/Remote/Api.php
```diff
@@ -444,16 +445,29 @@ public function updateActiveRecording($apiKey, array $record)
...
+        } else {
+            $user = $this->getUser();
+            $result = $this->getBackend()->get_current_recordings($user['userID']);
+        }
+       
         // no "last" activity existing
         if (count($result) == 0) {
             return $this->getErrorResult('No active recording.');
+        } else {
+            $timeSheetEntry=$this->getBackend()->timeSheet_get_data($result[0]);
```
simonschaufi: please trim all whitespace at the end of line

simonschaufi (changes requested):

libraries/Kimai/Remote/Api.php
```diff
@@ -444,16 +445,29 @@ public function updateActiveRecording($apiKey, array $record)
         if (empty($record)) {
             return $this->getErrorResult('Invalid record');
         }
+        $result = array();
```
simonschaufi: use short array syntax

libraries/Kimai/Remote/Api.php
```diff
@@ -444,16 +445,29 @@ public function updateActiveRecording($apiKey, array $record)
...
 
-        $user = $this->getUser();
-        $result = $this->getBackend()->get_current_recordings($user['userID']);
-
+        if (isset($record['timeEntryID'])) {
+            $result[]=$record['timeEntryID'];
```
simonschaufi: add whitespace around `=`

libraries/Kimai/Remote/Api.php
simonschaufi: add whitespace around `=`

simonschaufi (changes requested):

xakiru:
Thank you @simonschaufi , I made the requested changes.

simonschaufi (approved):

libraries/Kimai/Remote/Api.php
```diff
@@ -444,16 +445,29 @@ public function updateActiveRecording($apiKey, array $record)
...
+            $result[] = $record['timeEntryID'];
+        } else {
+            $user = $this->getUser();
+            $result = $this->getBackend()->get_current_recordings($user['userID']);
+        }
+       
         // no "last" activity existing
         if (count($result) == 0) {
             return $this->getErrorResult('No active recording.');
+        } else {
```
kevinpapst: Please remove the else, its not required

libraries/Kimai/Remote/Api.php
```diff
@@ -444,16 +445,29 @@ public function updateActiveRecording($apiKey, array $record)
...
+            $timeSheetEntry = $this->getBackend()->timeSheet_get_data($result[0]);
+        }
+
+        if (empty($timeSheetEntry)) {
+            return $this->getErrorResult('No active recording for the specified timeEntryID.');
         }
 
+
```
kevinpapst: Extra line is not required ;-)

kevinpapst:
Hi @xakiru ,
Thanks for your contributions, they are very welcome!
While looking through your PR I wondered why you changed the method `updateActiveRecording` to support passing in an entryId (this method is only meant to update currently a running entry).
Is the method `setTimesheetRecord` probably better fitted for your neeeds? It already supports setting a comment.

xakiru:
Hello Kevin,
Oh my bad I didn't see those methods, I was assuming that ActiveRecording are the only available methods. I just looked to the code, but it seems that setTimesheetRecord require too many fields for an update feature, don't you think ?

```
  // prepare data array - required values
        $data['userID'] = $user['userID'];
        $data['projectID'] = $record['projectId'];
        $data['activityID'] = $record['taskId'];
        $data['start'] = $in;
        $data['end'] = $out;
        $data['duration'] = $out - $in;
```

Basicly updating would require only 'id' to update.  
- stopping record: would be just by passing 'id' and 'end'
- starting record: would be just by passing 'id' and 'start'
- stop&comment: would be just by passing 'id',  'end' and 'comment'

And so on. Well, there is a way to make things work on my end, either by storing TimesheetRecord data or do an additional request to get TimesheetRecord data just before the updating request so I don't overwrite something edited from kimai panel. But this workaround would be the same for anyone who would use the api.

For the time being I'm going to roll back the change about updateActiveRecording, but the startRecord has to return the timeEntryID anyway otherwise it's really tricky to get the id using another request.

Thank you

simonschaufi:
This is how the mobile version does it: https://github.com/kimai/kimai-mobile/blob/master/kimai/kimai.mobile.js#L89-L103 It also fetches the active record and then updates it. Is this the way it should be? @kevinpapst are you more familiar with REST APIs how they usually work?

xakiru:
Well on the mobile it's using the stopRecord which doesn't have the possibility of adding a comment, in my case I want to stop the recording and update comment (when provided by user), so here what I'm going to do on stopping record:
getTimesheetRecord: get the recent data for a record
setTimesheetRecord: update the fields: end, comment (optional)

Thank you for the prompt replies guys 👍

xakiru:
After switching to use setTimesheetRecord, I had issues on updating records, after debugging I found out that the issue is coming from  the get_best_fitting_rate which return a null value. This should be fixed on my last change.

Thanks,

kevinpapst:
You could also adjust the method and make all fields optional during update?!
Then you don‘t need the previous fetch, which could potentially lead to race conditions.

This is no REST API in any means, just a basic API utilizing JSON ;-) so I wouldn‘t worry about specifications. And optional fields should not break existing apps (unless they relied on server side validation).

Viras-:
I strongly recommened to stay with at least a REST like approach - which means fetch the object -> update it -> store it.

xakiru:
Well, for the time being I will keep using the version where i'm doing fetch before any time_entry update, it's working fine. One more httprequest doesn't really matter after all.
So in this change I will stick to those bug fixes I made, just 2 small file changes:
- startRecord returns timeEntryID ( _instead of **nothing**_ )
- get_best_fitting_rate return 0 if no default rate is found ( _instead of **NULL**_ )
Thanks,

simonschaufi:
What i would suggest is the following:

Add a white listed array of possible attributes that can be updated in `updateActiveRecording` so that you don't make an if statement for all possible values and then cleanup some dirt there.

Then stopping a recording includes:
- Call `updateActiveRecording` with any comments made
- Call `stopRecord` if you know the currently running id or if you don't, then fetch it again by calling `getActiveRecording`.

xakiru:
Yea everything is working ok now, all I want is to push is the bug fixes.

xakiru:
Any news on this ?

simonschaufi:
I thought you would implement my suggestion but if you don't want, then I can merge it.

xakiru:
Yea since the api works as it is, no need to make changes I guess. The current made change is just a bug fixe.

simonschaufi:
@Viras- I didn't find your email address in the commit log so there is no other way to contact you than here. Could you get in touch with me via email: my github username [at] gmail [dot] com

Viras-:
Hi @simonschaufi I'm sorry I just saw your post and I just sent you an e-mail.

PR #1204, opened 2018-06-10
Convert timestamps to ints for proper rendering (Fix for #1032)

### LINKED ISSUE #1032: Time Format in Invoice
HI,

in the invoice we have date and time (start/end) of the work. this is fomated with [row.date;frm='dd.mm.yyyy'] for the date and [row.start;frm='hh:nn:ss'] for start and [row.end;frm='hh:nn:ss'] endtime. 

Now it comes to the stage that this is sometimes not working - have no idea why.
If we use the linux timestamp and recalculate this manually - this is working. so looks like that the entry in the database is correct.

Current Version : v1.1.0.1389
Upgrade to latest dev - Version: no Changes
Export from Kimai (not with invoice): correct format

This was working before always....
Have attached some screenshots.
Any Ideas ?

Many thanks
Dirk
with linux time - correct
word template
with formated time - red are wrong

mindhaq:
FIXES #1032

- `start` and `end` invoice data are converted explicitly to an integer before rendering.

Reason for this pull request:
- Tinybutstrong is used for rendering ODS invoices. It has a bug with certain timestamps, when they are passed as a string and not an integer. See https://github.com/Skrol29/tinybutstrong/issues/10


simonschaufi:
Thank you so much for helping us find the error! Your work is highly appreciated.

mindhaq:
Out of curiosity - my PHP skills are very rusty indeed - why did you change the `intval`call to a `int` cast?

simonschaufi:
because it's faster. see https://stackoverflow.com/questions/1912599/is-there-any-particular-difference-between-intval-and-casting-to-int-int-x#comment8750311_1912613

PR #1254, opened 2018-09-14
Translation refactoring

simonschaufi:
Refactor translation loading in order to make it possible to load language files from additional packages as well.

Reason for this pull request:
- Customer modifications of kimai

### DIFF
--- a/libraries/Kimai/Translation/Service.php
+++ b/libraries/Kimai/Translation/Service.php
@@ -26,6 +26,18 @@ class Kimai_Translation_Service
 {
 
     /**
+     * Load a translation data.
+     *
+     * @throws \Exception
+     */
+    public function load()
+    {
+        $language = Kimai_Registry::getConfig()->getLanguage();
+        $translationData = new Kimai_Translation_Data($language);
+        Kimai_Registry::setTranslation($translationData);
+    }
+
+    /**
      * Returns an array of all language codes.
      *
      * @return array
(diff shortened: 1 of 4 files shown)


libraries/Kimai/Translation/Service.php
```diff
@@ -25,6 +25,18 @@
...
 {
 
+    /**
+     * Load a translation data.
+     *
+     * @throws \Exception
+     */
+    public function load()
+    {
+        $language = Kimai_Registry::getConfig()->getLanguage();
```
kevinpapst, line 35: To be honest, I think this is a breach of the responsibility principle. 
From an architectural point of view the translation service should neither know nor care about a global registry.

kevinpapst:
working on a paid extension?

simonschaufi:
> working on a paid extension?

jup. new admin configuration section which I provide as composer package with minimal adjustments to kimai core. That's why I want to load my own translation files instead of patching kimai core.

libraries/Kimai/Translation/Service.php
simonschaufi: any better idea? the way how it was before is definitely not better.

simonschaufi: I'm loading my own translations with that:

```php
$this->translationData = Kimai_Registry::getTranslation();
$this->translationData->addTranslations('en', __DIR__ . '/');
$this->translationData->addTranslations('de', __DIR__ . '/');
Kimai_Registry::setTranslation($this->translationData);
```

if you have a better idea, I'm open for feedback ;)

kevinpapst: So, you are doing it like it was done before, but you find the way before not better?
Makes not much sense to me 😈 
Yes my proposal stands: leave as it was before. There is absolutely no reason why the translation service should know about the registry. Only the registry is global and knows about all objects it handles. 
Don't be offended - but I don't really care about Kimai 1. The code is crap and won't get better by any refactoring. If you like to move stuff around: do so!

kevinpapst:
Good idea to prepare the core to supply all necessary functions you need. 
But what you wanted to achieve was possible before, wasn't it?

kevinpapst (approved):

PR #1255, opened 2018-09-14
Remove double slash when using WEBROOT

simonschaufi:


kevinpapst:
Adjusting the constant is better, instead of changing the places where it is used ;-)

simonschaufi:
What do you prefer? starting a deeper path with a slash our without? I find it strange to start with a slash.
Are there best practices?

https://softwareengineering.stackexchange.com/a/84815

TYPO3 is adding a slash at the end of every constant (so you can start without): https://github.com/TYPO3/TYPO3.CMS/blob/b158d1795fb8bfb97e92cce90e3fc2192c1f8443/typo3/sysext/core/Classes/Core/SystemEnvironmentBuilder.php#L218

and if you don't need it, you can just chop it off: https://github.com/TYPO3/TYPO3.CMS/blob/b158d1795fb8bfb97e92cce90e3fc2192c1f8443/typo3/sysext/core/Classes/Core/SystemEnvironmentBuilder.php#L289

kevinpapst:
1. I would always do it this way around, multiple soft reasons
2. I would never change it, what if someone used it in in a private extension?

Do as you like, just make sure nothing breaks ;-)

simonschaufi:
Adjusting the constant would definitely be a breaking change

PR #1303, opened 2019-10-26
Fix issue #1302

### LINKED ISSUE #1302: Rounding Issue since v1.3.3
Since v1.3.3 there is an issue with the rounding-function. If i add an entry manually like this
kimai_rounding_before

it will actually be stored like that
kimai_rounding_after

If the end time had been 20:29:59, it would have been rounded correctly, like expected.
I expect, that times should be forced to be rounded to a full 15 minutes, no matter if the entry is added manually or "recorded".

If have activated these 2 options:
kimai_rounding_option_1
kimai_rounding_option_2

simonschaufi:
FIXES #1302 

- Fix rounding issue introduced in release 1.3.3

The same rounding rules are used as kimai v2 now. Thank you @kevinpapst!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - 7.2
+# due to incompatible phpunit framework
+#  - 7.2
 
 #matrix:
 #  allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
 language: php
 
 php:
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
   - 7.2
 
-matrix:
-  allow_failures:
-    - php: 7.2
+#matrix:
+#  allow_failures:
+#    - php: 7.2
 
 before_script:
   - ./.travis.install.sh
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         { "name": "Simon Schaufelberger", "homepage": "https://github.com/simonschaufi" }
     ],
     "require": {
-        "php": ">=5.5.0",
+        "php": ">=5.6.0",
         "phpoffice/phpexcel": "1.8.*",
         "tecnickcom/tcpdf": "^6.2.12",
         "tinybutstrong/tinybutstrong": "^3.10",
--- a/extensions/ki_timesheets/processor.php
+++ b/extensions/ki_timesheets/processor.php
@@ -624,8 +624,8 @@ switch ($axAction) {
             $outDate = new Zend_Date($edit_out);
 
             $rounded = Kimai_Rounding::roundTimespan(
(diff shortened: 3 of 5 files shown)


hh-it-co:
Thanks for your investigation. I've tested your fix, but it does not work as expected. Now, it seems like the rules for rounding will completely be ignored, except for recorded entries. The rounding is not forced as before during manual processing. Entries with less than the rounding-time will be rounded down, no matter if rounding down is allowed or not.

simonschaufi:
Do you mean the new release? I've tested it really well and it works how I would expect it.

If rounding down is allowed, it rounds either up or down. If it is not allowed, it always rounds up.

hh-it-co:
I will test again with 1.3.4

simonschaufi:
if you need the release zip file, I can provide that.

hh-it-co:
No, thanks. I test it directly from the repo.

hh-it-co:
It is still the same: If i start recording and stop it immediately after that, i expect it to round up to 15min. But it rounds down to 0min, no matter if i enabled rounding down or not.

hh-it-co:
Maybe i should go one step back:
My goal is to settle every quarter of an hour started. I expect to achieve this by activating the two options maintained in the error-descrioption.
Maybe I am wrong with this expectation?

simonschaufi:
Yes, the "fixed" situation is that the rounding will only take affect if at least 15 min are over since both values will be rounded up to the next 15min and will be the same.

From your expectation the expected result should be like this:

`1:05 - 1:06 -> 1:15 - 1:30`

Right?

If you expect the following, then you should activate the round down:

`1:05 - 1:06 -> 1:00 - 1:15`

but well, even that would currently only result in the values to be the same since `1:06` will still be rounded down. The other thing that I could do is to always round the start down and the end up. Then you would get the expected result (`1:05 - 1:06 -> 1:00 - 1:15`) I guess.

hh-it-co:
No, i would expect it to be

1:05 - 1:06 -> 1:00 - 1:15

hh-it-co:
In v1.3.3 the behavior was exactly as i expected. The only error thing was that the hole time-block was moved a quarter hour further after manually editing.
What i get in v1.3.4 is this (the displaying-errors are problems from the capturing software):

kimai_rounding_1 3 4

hh-it-co:
This is, what v1.3.3 does (same settings without rounding down). Thats, what i would call the right behavior:

kimai_rounding_1 3 3

simonschaufi:
@hh-it-co Expected behavior `1:05 - 1:06 -> 1:00 - 1:15` now implemented. Please update to release 1.3.5 and allow round down.

hh-it-co:
Works like a charm. Thank you very much!

simonschaufi:
You are welcome. If you want to support my work, you can do so at https://paypal.me/simonschaufi

PR #1306, opened 2019-11-05
Dutch translation

1willem:
Reason for this pull request:
- Adding the Dutch language (.nl.xliff) to Kimai2.

I translated the language files in the folder translations, as attached below. I assume some more files need to be translated. Can you tell me which files also need translation (log in screen, the error message when typing only one or two characters in the search bar, ... ?) and what other changes I need to make to implement Dutch language into Kimai?

Thanks for guiding me.

Willem

[Dutch Translations.nl.zip](https://github.com/kimai/kimai/files/3808551/Dutch.Translations.nl.zip)


kevinpapst:
Thanks @1willem 
But this is the wrong repository (Kimai 1).
Can you pleas re-open the PR here: https://github.com/kevinpapst/kimai2/pulls

PR #1308, opened 2019-11-08
How to build the release zip

uli-heller:
Reason for this pull request:
- I wanted to use version 1.3.5, but didn't find a kimai_1.3.5.zip. So I described how to create it.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,43 @@
+RELEASE
+=======
+
+Versions up to 1.3.2 provide a file named "kimai_{version}.zip".
+For newer versions, you'll have to create this file manually.
+I've done it using this description on an ubuntu 18.04 system.
+
+Install Various Dependencies
+----------------------------
+
+```
+sudo apt-get install -y ant
+sudo apt-get install -y composer
+sudo apt-get install -y php7.2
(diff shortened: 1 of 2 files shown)


simonschaufi (changes requested):
The changes in build.xml are valid but the kimai repo shouldn't be used for documentation. That should be done on the www.kimai.org repo and somewhere you will also find it in the v1 folder

uli-heller:
OK, I understand this, I'll modify the pull request. But why is there a file named INSTALL.md? It doesn't work unless you have a zip file as described in RELEASE.md (... which is going do disappear).

uli-heller:
Please note:  The description INSTALL.md is already included in www.kimai.org, file _v1/team.md

simonschaufi:
You don't necessarily need the zip file. The git repo and composer install will almost do the same. The zip file is just a bit smaller as i exclude some folders

uli-heller:
Am 2019-11-09 03:14, schrieb Simon Schaufelberger:

> You don't necessarily need the zip file. The git repo and composer 
> install will almost do the same.

Thx, but this means: additional dependencies for the installation. My 
kimai server doesn't have composer installed at the moment, for 
example...

PR #1309, opened 2019-11-08
Reduce size of release zip

uli-heller:
- Ignore some files when packing the ZIP

Reason for this pull request:
- Without this change, the ZIP has a size of 63M, with the change, the size is roughly 34M


simonschaufi:
@uli-heller what is the base branch for your PR? There are changes included which shouldn't be there.

uli-heller:
Somehow, I was not able to fix this pull request. So now there is a new one...

PR #1312, opened 2019-12-16
Grouped entries

simonschaufi:
Add the ability to group timesheet entries for export

### DIFF
--- a/extensions/ki_export/floaters.php
+++ b/extensions/ki_export/floaters.php
@@ -33,6 +33,7 @@ switch ($axAction) {
             'download_pdf' => 1,
             'customer_new_page' => 0,
             'reverse_order' => 0,
+            'grouped_entries' => 0,
             'pdf_format' => 'export_pdf',
             'time_type' => 'dec_time'
         ];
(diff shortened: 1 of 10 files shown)


simonschaufi:
@kevinpapst any opinion? A friend of mine requested this feature and I coded it for a :cake: in exchange ;)

kevinpapst:
The feature itself looks interesting and the query for sure is impressive ;-)

But afaik this kind of query is problematic, as this result might be "unexpected".  

One problem is AFAIK that you use non-aggregated columns in the group by. I wasn't aware of that in the past as well, but some database engines will not even execute such selects (MySQL will and silently return arbitrary data). The SQL standard does not allow such queries for a reason, but MySQL decided to allow them nevertheless.

Additionally you select other columns together with these grouped results (eg statusID and budget) and those are arbitrarily selected by MySQL (there is just one column result from X grouped rows). Those are picked more or less randomly from the all grouped rows. 

Hard to explain... the internet can explain it better 😁 you could search for something like this to start: https://stackoverflow.com/questions/3168042/can-i-use-non-aggregate-columns-with-group-by

Issue #1327, opened 2020-04-10
[defaultTimezone] not working


nwinesett:
#### Expected behaviour

1) In Admin Panel | Advanced, set America/New York for the field entitled: time zone for new users and customers 
2) Click Save and page reloads
3) Admin Panel | Customers | add customer
4) Dialog box opens but default customer is:  Africa/Abdijan

#### Actual behaviour

default customer is:  Africa/Abdijan

#### Steps to reproduce

See above

#### Reproducible on demo.kimai.org? [yes|no]

Not available anymore :(  ... I still love the speed and layout of v1

#### Kimai version [for example: dev 1.1.0.1389|github master|github develop]

v1.3.5.1400

#### Relevant log entries from log file (debug tab or temporary/logfile.txt)

None

simonschaufi:
As kimai 1 is not developed further, I'm only doing payed support. If you want the bug to be fixed, please sponsor me.

jimlucas:
I still prefer to use the older 1.3.x too.

kevinpapst:
@jimlucas then please keep on contributing and keep it alive! I am sure there are quite a lot of installation of v1 out there. 

If you ever feel as you want to contribute some ideas how v2 could be improved, please open feature requests for it.

nwinesett:
@jimlucas 
So cool!  I've copied your change to my installation.  I've tested it from both the admin panel and the timesheet lower lists and it works great in both scenarios.  Thanks for the commit.  I vote to update the master.
 
> I still prefer to use the older 1.3.x too.

@kevinpapst 
I've installed a test installation of v2 and tested it briefly on my site, but v1 is just so much quicker to use for me and everything is on one page.  I don't know v2 well enough to comment there, but I will keep it up to do an look at it from time to time.

I will tell you that in v1, my favorite feature is inline editing for the descriptions.  I can go through my work day without ever having to load a new page by simply starting a timer and selecting my radio buttons.  I've set up a customer named "New Client" and a project named "New Project" which serve as quick entries if somebody new calls.  I can just fill in a quick description and add the new customer and new project later.  No clicking or page loads to take me off my main screen.

I have some new good ideas (e.g. batch archiving - i.e. unchecking visibility) for v1 but I have to learn some coding to contribute better.  

Finally, this is long overdue, but a big thank you to the contributors of v1.  I know it's missing some features that you would get in a commercial product like review/approval workflow, but on the other hand, those commercial products don't touch the fast workflow in v1.

[...]

Issue #1326, opened 2020-03-21
If the language is "no-bo", creation of new customers isn't possible.


steinbitglis:
#### Expected behaviour
Creation of new customers should work as normal.

#### Actual behaviour
No panel for creating new customers will show up either in the admin panel, nor the front page.

#### Steps to reproduce
Go to the admin panel
Change the language to no-bo (Norwegian Bokmål)

#### Reproducible on demo.kimai.org? [yes|no]
No (it's Kimai 2)

#### Kimai version [for example: dev 1.1.0.1389|github master|github develop]
1.3.2

kevinpapst:
Please create this again in https://github.com/kevinpapst/kimai2/issues

simonschaufi:
@kevinpapst i believe he is talking about Kimai 1 as he mentioned the version

kevinpapst:
Hm...

> No (it's Kimai 2)
 
Now that I read it again ... maybe you are right and I misunderstood the sentence.

jimlucas:
March 21, 2020 7:36 AM, "Fredrik Ludvigsen" <notifications@github.com> wrote:

> Expected behaviour
> Creation of new customers should work as normal.
> 
> Actual behaviour
> No panel for creating new customers will show up either in the admin panel, nor the front page.

Fredrik,

I'm currently using v1.3.1.1394 and when adding a user I must enter the username to be added in the field to the left of the "Add User" button.  If you do not populate the username a javascript warning is thrown.

First, are you populating the username firld?
Second, is your browser possibly blockin JS pop-ups?


> 
> Steps to reproduce
> Go to the admin panel
> Change the language to no-bo (Norwegian Bokmål)
> 
> Reproducible on demo.kimai.org? [yes|no]
> No (it's Kimai 2)
> 
> Kimai version [for example: dev 1.1.0.1389|github master|github develop]
> 1.3.2
> 
> Relevant log entries from log file (debug tab or temporary/logfile.txt)
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.



--
Jim Lucas
CMS Webservices
5414085189

jimlucas:
First, no-bo is not recognized by the Zend Framework (at least the one being included in this version of kimai) as a valid locale.  And the Zend Framework (https://github.com/zendframework/zf1) project was EOL as of 2016-09-28.  So we can't expect any updates from that repo.

[...]

PR #1328, opened 2020-04-11
Fix issue #1327 and #1326

jimlucas:
FIXES #

- Allows for default timezone to be auto selected when adding new customer

Reason for this pull request:
- Ability for the default timezone to be selected by the system when adding a new customer was broken when related code was refactored to be nested under a sub array called 'customer' rather than each key/value assigned separately

### DIFF
--- a/core/floaters.php
+++ b/core/floaters.php
@@ -124,7 +124,11 @@ switch ($axAction) {
             $view->assign('id', 0);
         }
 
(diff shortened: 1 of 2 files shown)


nwinesett (approved):
I have uploaded to my installation and it works.

jimlucas:
I need to break these up into separate requests.  I'll be back.

PR #1338, opened 2020-08-28
fix warnings and deprecation notices when using php 7.4

ulrich-berkmueller:
* replaces `while each` loops with `foreach` loops
* fixes order of `$glue` and `$pieces` of `implode` function calls (<https://www.php.net/manual/de/function.implode.php>)
* fixes accessing chars of a string (using square instead of curly braces)
* use `get_magic_quotes_gpc` only if PHP version is less than 5.4.0 (see <https://www.php.net/manual/de/function.get-magic-quotes-gpc.php>)

### DIFF
--- a/libraries/phpclasses/ultimatemysql/mysql.class.php
+++ b/libraries/phpclasses/ultimatemysql/mysql.class.php
@@ -1588,9 +1588,6 @@ class MySQL
 				if (strlen($value) == 0) {
 					$return_value = "NULL";
 				} else {
-					if (version_compare(PHP_VERSION, "5.4.0", '<') && get_magic_quotes_gpc()) {
-						$value = stripslashes($value);
-					}
 					$return_value = "'" . str_replace("'", "''", $value) . "'";
 				}
 				break;
--- a/libraries/phpclasses/ultimatemysql/mysql.class.php
+++ b/libraries/phpclasses/ultimatemysql/mysql.class.php
@@ -1588,7 +1588,7 @@ class MySQL
 				if (strlen($value) == 0) {
 					$return_value = "NULL";
 				} else {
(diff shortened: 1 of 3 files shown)


libraries/phpclasses/ultimatemysql/mysql.class.php
```diff
@@ -1588,7 +1588,7 @@ static public function SQLValue($value, $datatype = self::SQLVALUE_TEXT) {
 				if (strlen($value) == 0) {
 					$return_value = "NULL";
 				} else {
-					if (get_magic_quotes_gpc()) {
+					if (version_compare(PHP_VERSION, "5.4.0", '<') && get_magic_quotes_gpc()) {
```
kevinpapst: creepy 😁

kevinpapst (approved):
looks good to me, but you know that this software is not really maintained any longer?

libraries/phpclasses/ultimatemysql/mysql.class.php
ulrich-berkmueller: Yes I know, but it still works for me since I only use a small subset of features.
Just wanted to contribute back a little bit. :-)

ulrich-berkmueller: I can also do a version compare for <7.4. Sine 5.4 `false` is always returned, since 7.4 there's also a deprecation notice.

kevinpapst: This code can imo be removed completely. 
Who nowadays uses a PHP < 5.4 (EOL 3 Sep 2015) has to live with the consequences...

kevinpapst (approved):
Thanks 👍

ulrich-berkmueller:
You're welcome!

